PR #6, opened 2020-03-23
Fix some minor issues in documentation

btoll:
https://www.grammar-monster.com/easily_confused/its_its.htm
https://www.grammarly.com/blog/whos-whose/

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -62,7 +62,10 @@ jobs:
     - Echo
 ```
 
-Here we have a mule config file that lists one task and one job. The tasks field is where you document task configurations that you will need to execute your jobs. It takes a list of maps (dicts in python) that are used to initialize your tasks. One key thing to keep in mind here is that every task configuration must have a task field set. This is used by mule to loop through the task you wish to be initialize. Every other field is a configuration for your task. In this example, we have one task called Echo, which simply prints out "Message for the console".
+Above is a mule config file that lists one task and one job.
+
+- The tasks field is to document task configurations that are needed to execute jobs. It takes a list of maps (dicts in python) that are used to initialize your tasks -- keep in mind that every task configuration must have a task field set -- and is used by mule to loop through the task you wish to initialize.
+- Every other field is a configuration for your task. In this example, we have one task called Echo, which simply prints out "Message for the console".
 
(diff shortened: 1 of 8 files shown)


docs/getting_started.md
```diff
@@ -62,7 +62,7 @@ jobs:
     - Echo
 ```
 
-Here we have a mule config file that lists one task and one job. The tasks field is where you document task configurations that you will need to execute your jobs. It takes a list of maps (dicts in python) that are used to initialize your tasks. One key thing to keep in mind here is that every task configuration must have a task field set. This is used by mule to loop up the task you wish to be initialize. Every other field is a configuration for your task. In this example we have one task called Echo, which simply prints out "Message for the console".
+Here we have a mule config file that lists one task and one job. The tasks field is where you document task configurations that you will need to execute your jobs. It takes a list of maps (dicts in python) that are used to initialize your tasks. One key thing to keep in mind here is that every task configuration must have a task field set. This is used by mule to loop through the task you wish to be initialize. Every other field is a configuration for your task. In this example, we have one task called Echo, which simply prints out "Message for the console".
```
btoll: Here I changed "loop up" to "loop through".

onetechnical:
```suggestion
Here we have a mule config file that lists one task and one job. The tasks field is where you document task configurations that you will need to execute your jobs. It takes a list of maps (dicts in python) that are used to initialize your tasks. One key thing to keep in mind here is that every task configuration must have a task field set. This is used by mule to add the task you wish to be initialized. Every other field is a configuration for your task. In this example, we have one task called Echo, which simply prints out "Message for the console".
```

bricerisingalgorand (dismissed):
I approve of this message

docs/getting_started.md
onetechnical: This sentence is awkward...I think the idea was to add the task (hence the above suggestion). Regardless, the "initialize" needs to be modified, whichever way it goes.

onetechnical (changes requested):
Ick, a grammar PR! Grammared the grammar changes.

onetechnical (approved):

bricerisingalgorand (approved):
I still approve of this message

PR #7, opened 2020-03-25
Brice/support env vars docker

bricerisingalgorand:
This refactors our mule tasks that require docker containers to all use the syntax, which includes env vars:

```yaml
  - task: docker.Make
    docker:
      image: algorand/go-algorand-ci-linux
      version: '{{ docker.Version.outputs.version }}'
      workDir: /go/src/github.com/algorand/go-algorand
      env:
        - env_var=env_var_val
    target: ci-build
    dependencies: shell.docker.Ensure
```

Some changes to how we evaluate parameterized fields and requirements of types in required/optional fields were made here to support the new feature.


mule/task/__init__.py
```diff
@@ -22,14 +25,38 @@ def __init__(self, args):
...
+                    if type(field_value) is str:
+                        fieldDict[field] = pystache.render(field_value, job_fields)
+                    if type(field_value) is list:
+                        for element_index, element in enumerate(field_value):
+                            if type(element) is str:
+                                fieldDict[field][element_index] = pystache.render(element, job_fields)
+                            elif type(element) is dict:
+                                fieldDicts.append(element)
+                    elif type(field_value) is dict:
+                        fieldDicts.append(fieldDict[field])
```
egieseke, line 59: Should there be an else condition to print a warning that the type is not supported?

egieseke (approved):
looks good.

btoll (approved):

mule/task/__init__.py
bricerisingalgorand: Users can include any type they need to, but only strings can be templates since there would be no other way to represent a mustache template in yaml. Any string should be getting evaluated by this method, unless I missed something

PR #8, opened 2020-03-26
Brice/fix docker run bug

bricerisingalgorand:
Fix bug I accidentally introduced in the last PR to introduce docker env vars


egieseke (approved):
looks good.

PR #9, opened 2020-03-31
Brice/fix order of docker ensure

bricerisingalgorand:
Fixed the order for the docker ensure process. Now it checks image locally -> image in docker up -> builds image. I also did some gold plating around the messages being printed and refactored the kill function to only kill the container if it actually exists


egieseke (approved):
Looks good.

PR #12, opened 2020-03-31
Brice/fix env var evaluation

bricerisingalgorand:
Now failed env var evaluations will produce an empty string, rather than the bash style env var syntax. Also updated IDockerTask to validate that all env vars are formatted correctly


egieseke (approved):
OK, looks fine.

PR #16, opened 2020-04-15
Support environment variables in task definitions

btoll:
The most important bit about this PR is that it moves the parsing of the environment variables to the first operation rather than after it is validated.

While it certainly will perform better to isolate the job and its tasks first and then only parse those subsections of the document rather than the entire document, I found that calling the parsing functions more than once leads to problems, the most significant of which is that not every environment variable will be parsed.

For example, in order for an env var to be parsed, it must have its tag set to the one passed to the yaml module when calling both `add_implicit_resolver` and `add_constructor`.  In this case, it's "!path".  The PyYAML's `ScalarNode` type should have its default tag swapped out for this one, resulting in a dict such as `{ "tag": "!path", "value": "$HOME is where the hate is" }`.

However, this doesn't happen when `mule`'s env var parsing function is called a second time.  The `tag` is set (or not) from the default in `composer.py#compose_scalar_node` (the default tag value is `tag:yaml.org,2002:str`).  Again, if the "!path" key to the resolver's matrix of regular expressions isn't set correctly here, then the environment variable in the scalar node that is being examined will never be parsed (i.e., the constructor passed to the yaml module's `add_constructor` function is never called).

 I spent quite a bit of time stepping through the PyYAML library until I finally decided that it's better to parse the entire document up front.


mule/util/yaml/env_var_loader.py
```diff
@@ -4,7 +4,8 @@
 from mule.error import messages
 from termcolor import cprint
 
-env_var_path_matcher = re.compile(r'.*\$\{([^}^{]+)\}.*')
+env_var_path_matcher = re.compile(r'.*\$\{?([^}^{]+)\}?.*')
```
btoll, line 7: I made the brackets optional so that the yaml parser doesn't choke on lines like the following:
```
volume: [ ${HOME}/go/bin:go ]
```

mule/driver.py
```diff
@@ -11,7 +11,9 @@
 def main():
     args = mule.parser.parseArgs()
     try:
-        jobs_config, task_configs = validator.getValidatedMuleYaml(args.file)
+        mule_config = file_util.readYamlFile(args.file)
+        parsed_mule_config = yaml_util.readYamlWithEnvVars(yaml.dump(mule_config))
```
bricerisingalgorand, line 15: I think we shouldn't do env var evaluations for the entire file here, because doing this will show warnings for tasks/jobs that aren't being executed. This is why I was suggesting evaluating task env vars during task instantiation. That way we are only evaluating env vars for tasks we are planning to execute

btoll: I first looked at doing it that way, but I ran into issues as I described above that I can go into more detail about on a call, if you'd like.

What warnings are you referring to?

bricerisingalgorand (approved):
looks good to me!

egieseke (approved):
looks good.

PR #18, opened 2020-04-27
Brice/docker shell config option

bricerisingalgorand:
This is helpful so that users can use containers that don't come with bash installed (alpine)


egieseke (approved):
Looks good.

btoll (approved):

PR #20, opened 2020-04-29
Remove "outputDir" from download file(s) required fields list

btoll:
"outputDir" is an optional field that defaults to ".", so I removed it from the `required_fields` list so it can truly be optional.


egieseke (approved):
good update.

onetechnical (approved):

PR #21, opened 2020-05-06
Don't default outputDir to None

btoll:
This will cause an exception b/c a helper method in `download_file` is expecting a string not a None value.


egieseke (approved):
looks ok.

onetechnical (approved):
👍

PR #22, opened 2020-05-11
Added new `BucketCopy` class for s3

btoll:
```
tasks:
  # entire s3 bucket -> local
- task: s3.BucketCopy
  name: copy1
  src: s3://ben-test-2.0.3/foo/bar/
  dest: .

  # entire s3 bucket -> local
- task: s3.BucketCopy
  name: copy2
  src: s3://ben-test-2.0.3/foo/bar
  dest: .

  # s3 file -> local
- task: s3.BucketCopy
  name: copy3
  src: s3://ben-test-2.0.3/foo/bar/1.out
  dest: .

  # s3 glob -> local
- task: s3.BucketCopy
  name: copy4
  src: s3://ben-test-2.0.3/foo/bar/*.out
  dest: daffy

  # local glob -> s3
- task: s3.BucketCopy
  name: copy5
  src: daffy/*.out
  dest: s3://ben-test-2.0.3/foo/bar/

  # local filename -> s3
- task: s3.BucketCopy
  name: copy6
  src: daffy/3.out
  dest: s3://ben-test-2.0.3/foo/bar/

  # entire local dir -> s3
- task: s3.BucketCopy
  name: copy7
  src: daffy
  dest: s3://ben-test-2.0.3/foo/bar/

  # s3 glob -> local
- task: s3.BucketCopy
  name: copy8
  src: ./*.out
  dest: s3://ben-test-2.0.3/foo/bar/

  # local -> local
- task: s3.BucketCopy
  name: error
  src: ./*.out
  dest: test

jobs:
  copy1:
    tasks:
      - s3.BucketCopy.copy1
      - s3.BucketCopy.copy2
      - s3.BucketCopy.copy3
      - s3.BucketCopy.copy4

  copy2:
    tasks:
      - s3.BucketCopy.copy5
      - s3.BucketCopy.copy6
      - s3.BucketCopy.copy7
      - s3.BucketCopy.copy8

  error:
    tasks:
      - s3.BucketCopy.error
```


bricerisingalgorand:
Would it be possible to use the syntax the aws cli uses to generically copy either s3 -> local, local -> s3 and s3 -> s3?

btoll:
> Would it be possible to use the syntax the aws cli uses to generically copy either s3 -> local, local -> s3 and s3 -> s3?

That's a good suggestion, I'll make it so.

btoll:
I've been thinking about this, and what do you think of either 1) using UploadFiles task to copy local -> s3 or 2) having a method in the BucketCopy class to delegate to that class (and same in for DownloadFiles when going s3 -> local).

If I'm misunderstanding your question, please elaborate.

bricerisingalgorand:
Hmm I'm not sure why we wouldn't want to go ahead and have this copy operation be the same as the s3 copy operation. All we'd need to add to write it that way is something to parse the `s3://${BUCKET_NAME}/${OBJECT_NAME}` string. The rest of the utility functions we'd need are there pretty much

If we stick to only copying files between s3 buckets, maybe we need to come up with a more specific name for the task. What do you guys think @onetechnical @egieseke ?

btoll:
Sounds like you prefer #2, I'll implement that.

mule/task/s3/__init__.py
```diff
@@ -104,3 +108,68 @@ def execute(self, job_context):
...
+                # To get the glob, join the last two tuple elements, i.e., ('', 'bar, 'foo') => 'bar/foo'
+                s3_util.download_files(src_bucket, prefix, suffix, '/'.join((dest_bucket, dest_prefix)))
+            else:
+                # Uploading local -> s3.
+                # To get the glob, join the last two tuple elements, i.e., ('', 'bar, '*.out') => 'bar/*.out'
+
+                # Handle cases where just the base directory is given (with or without trailing forward slash).
+                if not src_prefix:
+                    src_prefix = '*'
+                s3_util.upload_files('/'.join((src_bucket, src_prefix)), dest_bucket, dest_prefix)
```
bricerisingalgorand, line 167: I'm confused, this is the conditional where src is local right? Why are we calling this src_bucket and src_prefix? I think we need to refactor this some. We can refactor this so that we have a method determining whether a src or dir is remote/local, then another that parses a remote src/dir. That way variable names can be named appropriately. The way it is now is kind of hard to read

mule/task/s3/__init__.py
```diff
@@ -104,3 +108,68 @@ def execute(self, job_context):
...
+
+    # https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html
+    bucketRe = re.compile(r'^(s3://)?([a-z0-9-.]*)\/?(.*)?$')
+
+    def __init__(self, args):
+        super().__init__(args)
+        self.src = args['src']
+        self.dest = args['dest']
+
+    def copy(self):
```
bricerisingalgorand, line 126: This method is doing a bit too much. Instead of all these conditionals, we can make some functions in s3_util that can make them simpler. That should help a lot with readability

mule/task/s3/__init__.py
```diff
@@ -104,3 +108,68 @@ def execute(self, job_context):
...
+                # Downloading s3 -> local.
+
+                prefix = src_prefix
+                suffix = ''
+
+                if src_prefix[-1] != '/':
+                    path = pathlib.Path(prefix)
+                    # We need to check for suffixes here, i.e. .tar.gz, which would return [.tar, .gz].
+                    # If we just check for extension, we'll only get .gz, which may not be what we want.
+                    suffix = "".join(path.suffixes) if len(path.suffixes) else ''
```
bricerisingalgorand, line 144: Does os.path.basename not work here? I'm a little confused because if someone gives the absolute path to the file in s3, can we not just make the prefix=the absolute path?

mule/task/s3/__init__.py
```diff
@@ -104,3 +108,68 @@ def execute(self, job_context):
...
+
+                    # The current download_files function does not support globs, so we need to futz a bit
```
bricerisingalgorand, line 146: I think @egieseke couldn't use globs because s3 doesn't support it. This makes sense because s3's not actually a file system. Implementing glob here will be a lot more work than this and it's kind of out of scope of the effort. It's ok to not support globs for s3 paths as long as we document it in our task documentation (which might be missing this suite of tasks, but we're getting there 👍 )

btoll: It was just an explanatory comment, not a criticism.

mule/task/s3/__init__.py
btoll: I think having it here in the class is ok for now.  Currently, this class is the only one that needs this specific behavior.  If we find that another class needs it, let's pull it out then.

The utility file is becoming a dumping ground for every function, shared or not, among the classes, which I think affects readability as well.

mule/task/s3/__init__.py
btoll: I'm collecting the bits that need to be passed to the helper methods.  The pathlib library is pretty cool, it can easily determine the suffix(es)/extensions of a filename, which we need to determine to pass to our helper methods.

Using basename works, of course, but here we need to determine just the suffixes, not the entire filename.

Keep in mind that this is supporting everything that the upload and download classes.

mule/task/s3/__init__.py
btoll: The regex on lines 127 and 128 is already determining the src and dest bits that are needed to be sussed out to be passed to the helper functions based upon the operation that is needed.

mule/task/s3/__init__.py
bricerisingalgorand: I know, what I mean is that I don't believe we need to support a '*' character in s3 paths all together. That's what line 155 is doing right?

bricerisingalgorand (approved):
@onetechnical @egieseke We need to have a conversation about how we want to handle glob expressions for s3 file paths

onetechnical (approved):
I think this looks okay (though I didn't run it). We should discuss S3 globs.

PR #24, opened 2020-05-14
Suggestion to just have one upload and download class

btoll:
The download class will delegate to the proper function depending upon its inputs, and the upload class will always use one path to take advantage of globs.


mule/task/s3/__init__.py
```diff
@@ -47,39 +27,28 @@ def execute(self, job_context):
...
         self.bucketName = args['bucketName']
         self.prefix = args['prefix']
         self.suffix = args['suffix'] if 'suffix' in args else None
+        self.fileName = args['fileName'] if 'suffix' in args else None
+        self.objectName = args['objectName'] if 'suffix' in args else None
         self.outputDir = args['outputDir'] if 'outputDir' in args else '.'
 
     def download_files(self):
-        s3_util.download_files(self.bucketName, self.prefix, self.suffix, self.outputDir)
+        if self.objectName or self.filename:
```
egieseke: Is this a potential error condition if the output file name is provided, but the object name is not?

btoll: I just tested and no error, it looks ok.

btoll: I just tested and no error, it looks ok.

egieseke: Should you raise an error in this case? since the assumption is that there is one and only one object to download into the specified file name.

btoll: There are a couple of ways to download files:
1. Specify a file either as an object name or a filename.
1. If there is no #1, a prefix and/or suffix is expected.

I think you're right, there should be an exception raised if none of those conditions are met.  I'll add one.

mule/task/s3/__init__.py
```diff
@@ -46,40 +25,23 @@ def execute(self, job_context):
...
-    required_fields = [
-        'bucketName',
-        'prefix'
+        'bucketName'
     ]
 
     def __init__(self, args):
         super().__init__(args)
         self.bucketName = args['bucketName']
         self.prefix = args['prefix']
```
bricerisingalgorand, line 36: it looks like prefix was removed from the required fields list

mule/task/s3/__init__.py
```diff
@@ -46,40 +25,23 @@ def execute(self, job_context):
 
 class DownloadFile(ITask):
```
bricerisingalgorand: I think if we're going to make this change, this needs to be renamed to just Download possibly. DownloadFile doesn't make sense if it will download 1 or many files

mule/task/s3/__init__.py
btoll: That was a mistake, correcting.

mule/task/s3/__init__.py
```diff
@@ -46,40 +25,23 @@ def execute(self, job_context):
...
         self.bucketName = args['bucketName']
         self.prefix = args['prefix']
         self.suffix = args['suffix'] if 'suffix' in args else None
+        self.fileName = args['fileName'] if 'fileName' in args else None
+        self.objectName = args['objectName'] if 'objectName' in args else None
         self.outputDir = args['outputDir'] if 'outputDir' in args else '.'
 
     def download_files(self):
-        s3_util.download_files(self.bucketName, self.prefix, self.suffix, self.outputDir)
+        if self.objectName or self.fileName:
```
bricerisingalgorand: We'll need some handling around when a fileName is set but no objectName

mule/util/s3_util.py
```diff
@@ -89,10 +89,11 @@ def download_file(bucket_name: str, object_name: str, output_dir: str = ".", fil
     return True
 
 
-def download_files(bucket_name: str, prefix: object = "", suffix: object = "", output_dir: str = ".") -> bool:
+def download_files(bucket_name: str, object_name: str, prefix: object = "", suffix: object = "", output_dir: str = ".") -> bool:
```
bricerisingalgorand: I'm confused, isn't this a syntax error? The header doesn't match the call from the s3 mule task. This param isn't necessary right?

mule/task/s3/__init__.py
```diff
@@ -1,28 +1,7 @@
 from mule.task import ITask
 from mule.util import s3_util
 
-
 class UploadFile(ITask):
```
bricerisingalgorand: Same comment as the DownloadFile name, but this would probably need to be Upload instead also

mule/task/s3/__init__.py
btoll: Will change.

mule/task/s3/__init__.py
btoll: Yes, I'm adding that now.

mule/task/s3/__init__.py
btoll: Will change.

mule/util/s3_util.py
btoll: Probably.  Initially, this PR was a draft and intended to open a discussion on this.  I threw this into the signature just to have something there, but I'll look more closely at it now.

mule/task/s3/__init__.py
btoll: Actually, it looks as though this should be an `or` situation.  The `objectName` is the full key name of the object to download, so both options wouldn't need to be set.

mule/util/s3_util.py
btoll: Fixed

algojohnlee:
Looks good to me, yay simpler

onetechnical (dismissed):
Looks good to me, yay simpler

mule/task/s3/__init__.py
bricerisingalgorand: but what about when someone uses these configurations

```yaml
- task: s3.Download
  bucket: mybucket
  prefix: myprefix
  fileName: test.jpg
```

s3_util.download_file will get called with an objectName of none and throw an exception right?

mule/task/s3/__init__.py
egieseke: Did you want to add an exception in case the object name was not specified?

mule/task/s3/__init__.py
```diff
@@ -75,11 +34,18 @@ def __init__(self, args):
...
         self.outputDir = args['outputDir'] if 'outputDir' in args else '.'
 
     def download_files(self):
-        s3_util.download_files(self.bucketName, self.prefix, self.suffix, self.outputDir)
+        if self.objectName or self.fileName:
+            s3_util.download_file(self.bucketName, self.objectName, self.outputDir, self.fileName)
+        elif self.suffix:
+            s3_util.download_files(self.bucketName, self.prefix, self.suffix, self.outputDir)
+        else:
+            raise MissingOption('objectName or fileName must be specified')
```
bricerisingalgorand: objectName and fileName aren't listed as required fields, so why are you raising this exception here? Also you're not handling the basic happy path situation of just the required fields being set

bricerisingalgorand: This else could probably handle the basic happy path case.

bricerisingalgorand (changes requested):

mule/task/s3/__init__.py
btoll: They can't be listed as required fields because files can be downloaded by only specifying `prefix` and `suffix`.  Also, isn't `suffix` necessary as well if not specifying either `fileName` or `objectName`?

I think what you're hitting on is that this API is a bit confusing and that's a legitimate criticism, since files can be downloaded in more than one way.

mule/task/s3/__init__.py
btoll: I see what you're saying.  Yes, I'll fix that.

mule/task/s3/__init__.py
btoll: I don't think so, because a `fileName` is the only field needed to download a file (along with `prefix`, which is required).

If both were specified, which would win?

btoll: Ah, nevermind that last comment.  I misunderstood the filename to be the name of the file that was wanted to be downloaded.

I'll fix that.

egieseke (approved):
Looks good.

PR #26, opened 2020-05-15
Brice/impelement stash

bricerisingalgorand:
This implements a concept of a stash into the mule cli. It works similarly to jenkin's stash function. It tars up all files in a glob expression and throws it to s3 on a specified id. Then downloads that tarfile based on that id later on when it is needed. All glob expressions are done with respect to the current working directory, and that current working directory is where the tar is decompressed on unstash as well.


btoll:
I could see a situation where someone would forget the location from which they had stashed, and then when they unstash the extracted tarball litters the cwd.  Not so bad if the archive is all contained in a directory, but fairly troublesome if it throws individual files into the cwd.

I like the idea of stash, but not so sure about unstash.  What about just using the download file task to get the stash and then the user can extract where they wish?  Or having an option in the unstash class to download but not extract?

btoll:
I also like the idea of having sensible defaults (like the cwd) for where to create/extract the tarball, but the option to specify would be nice.

bricerisingalgorand:
The stash maintains the folder structure of the project when the stash is called, so it shouldn't drop files into the wrong directory when it's unstashed. I can add the ability to set cwd, but I think that would get a little confusing because the glob expression for stashing will need to be with respect to that cwd and it'll maintain the folder structure from that directory, so using a different cwd on the unstash may not end up being useful. The reason why I wrote it without that setting was to get feature parity with jenkin's stash feature that is pretty popular https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#stash-stash-some-files-to-be-used-later-in-the-build

mule/task/stash/__init__.py
```diff
@@ -0,0 +1,48 @@
+from mule.task import ITask
+from mule.util import s3_util
+from mule.util import time_util
+from mule.util import file_util
+
+class Stash(ITask):
+    required_fields = [
+        'bucketName',
+        'stash_id',
```
egieseke: Starting to mix task argument naming with camel case and underscores.  We should be consistent.

egieseke (dismissed):
Looks good, except for mixed styles of task arguments.

btoll (dismissed):

PR #27, opened 2020-05-15
Treat "prefix" config as a directory/folder on s3

btoll:
```
- task: s3.UploadFiles
    name: upload_my_files
    bucketName: ben-test-2.0.3
    prefix: test
    globSpec:
      - ./*.out
```
    
This will upload files to `s3:ben-test-2.0.3/test`, i.e.,
    
    ben-test-2.0.3/test/1.out
    ben-test-2.0.3/test/2.out
    ben-test-2.0.3/test/3.out
    
The current behavior uploads these as:
    
    ben-test-2.0.3/test1.out
    ben-test-2.0.3/test2.out
    ben-test-2.0.3/test3.out
    
I think this is confusing and not what is expected from the "prefix" field.
    
This article treats "prefix" as a synonym of "folder" or "directory":
https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html

### DIFF
--- a/mule/util/s3_util.py
+++ b/mule/util/s3_util.py
@@ -17,12 +17,19 @@ def _path_leaf(path: str) -> str:
     return tail or ntpath.basename(head)
 
 
-def upload_files(globspec: object, bucket_name: str, prefix = None) -> bool:
+def _get_object_name(file: str, preserveDirs: bool) -> str:
+    if preserveDirs:
+        return file.replace('./', '')
+
+    return os.path.basename(file)
+
+def upload_files(globspec: object, bucket_name: str, prefix = None, preserveDirs = False) -> bool:
     """
     Upload files using a list of globs to an S3 bucket.
     :param globspec: Glob to match local files
     :param bucket_name: Bucket name to upload to.
-    :param prefix: Prefix for object names placed in s3
+    :param prefix: Upload objects whose key starts with this prefix (optional).
+    :param preserveDirs: Preserve the local directory structure of the file(s) location on the server (optional).
     :return: True if successful, otherwise False.
     """
     response = True
@@ -34,7 +41,7 @@ def upload_files(globspec: object, bucket_name: str, prefix = None) -> bool:
         files = glob.glob(globspec, recursive=True)
         for file in files:
             if os.path.isfile(file):
-                object_name = os.path.basename(file)
+                object_name = _get_object_name(file, preserveDirs)
                 if prefix != None:
                     # Remove both "./" from the beginning and "/" from both sides, if present.
                     object_name = f"{prefix.lstrip('.').strip('/')}/{object_name}"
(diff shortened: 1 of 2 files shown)


mule/util/s3_util.py
```diff
@@ -34,10 +34,10 @@ def upload_files(globspec: object, bucket_name: str, prefix = None) -> bool:
         files = glob.glob(globspec, recursive=True)
         for file in files:
             if os.path.isfile(file):
-                file = file.replace('./', '')
-                object_name = file
+                object_name = os.path.basename(file)
```
bricerisingalgorand: This won't maintain the folder structure of the path returned by the glob expression. This may be a use case some people would want, but our stash task depends on the path being maintained. If we wan't to support uploading files without maintaining their folder structure, we need to make it optional

bricerisingalgorand:
I still don't love this idea since from my pov it removes features without providing much in return, but I'm ok with it if I get outvoted

mule/util/s3_util.py
btoll: In my opinion, it shouldn't maintain the folder structure, and I think that maintaining that structure goes against expectations.

It's confusing to want to upload the files in `/foo/bar/quux/*.sh` and have that directory structure maintained on the server (unless, of course,  the destination is`s3:mybucket/foo/bar/quux`).

btoll:
> I still don't love this idea since from my pov it removes features without providing much in return, but I'm ok with it if I get outvoted

Could you please link to some aws/boto3 documentation that supports your interpretation of `prefix`?  Everything I've seen uses it as a key prefix, not as a way to also rename the file/object.

In my opinion, what you're referring to as features are assumptions that are being made that go against well-established use cases and expectations.  If the Stash class need to maintain the directory structure, then it should have its own implementation of upload that does this instead of using the general upload function.

algojohnlee:
My vote - I think in general Ben's right on this one...if I do something like `aws s3 cp --recursive s3://mybucket/foo .` it will copy the contents of the foo bucket and not a file called foobar.txt at the root. Hence, for devs, it is probably more intuitive to treat it as a folder.

I do kind of like the hypothetical of being able to write custom object name prefixes, but I think we should do that in the object name, not the prefix in general.

onetechnical:
My vote - I think in general Ben's right on this one...if I do something like `aws s3 cp --recursive s3://mybucket/foo .` it will copy the contents of the foo bucket and not a file called foobar.txt at the root. Hence, for devs, it is probably more intuitive to treat it as a folder.

I do kind of like the hypothetical of being able to write custom object name prefixes, but I think we should do that in the object name, not the prefix in general.

mule/util/s3_util.py
bricerisingalgorand: I'm not sure I understand how flattening the folder structure is necessarily more intuitive. A lot of filenames are only meaningful within their respective folder structure. Also it's only a couple lines of code to make that conditional. Requiring end users to write wrapper scripts to do that themselves kind of recreates the problem this framework is trying to solve.

bricerisingalgorand:
Hey, you're not allowed to vote twice on two different accounts! @onetechnical @algojohnlee 

I would counter that argument by saying that there the aws cli is attempting to create an implementation similar to the posix cp. We're using language like Download and Upload, which doesn't come with the same implications.

btoll:
> I would counter that argument by saying that there the aws cli is attempting to create an implementation similar to the posix cp. We're using language like Download and Upload, which doesn't come with the same implications.

I really think we should be very careful about designing APIs that have side effects, which is what you're proposing.

Let's take a specific example.  I'm uploading files via a webpage, and I'm browsing to my files that are nested several directories deep.  When I click upload, I'm not expecting that the server is going to create that entire directory structure (essentially, mirror it) on the server.  It should be uploaded to where I specified, and I'm not expecting the implementation to have side effects that make another decision for me.

That's for a generic Upload/Download implementation.  I think that it's perfectly fine for the stash class to have its own upload implementation that will keep the directory structure intact, but that shouldn't be the generic case.  It's specific to a specific operation.  Also, it should be documented that the operation has side effects.

Same argument goes for using the `prefix` key for changing the filenames on the server.

mule/util/s3_util.py
btoll: It's intuitive because that's how upload operations work.  Could you link to examples where it works as you're describing?

Stash can have its own implementation of upload, there's no reason or need for end users to have to write a wrapper script or anything else.

btoll:
I'm reopening this since there seems to be more consensus around this.

bricerisingalgorand (approved):
looks good to me

egieseke (approved):
looks good

PR #28, opened 2020-05-19
Change stash to support multiple globspecs

bricerisingalgorand:
Now this field takes a list and supports a list of glob expressions


egieseke (approved):
Looks good.

PR #29, opened 2020-05-29
Brice/sdk pipeline

bricerisingalgorand:
This fixes a few bugs in the shell.docker tasks and adds a couple new features that are nice to have for the sdk pipelines. One that may be controversial is giving Echo the ability to run quietly. I did that so it could be used to store computed values in the pipeline. It pairs conveniently with the new "!timestamp" feature that's introduced in this PR. I also added the ability to pass in a docker network for docker tasks since that's necessary for our sdk pipelines.


btoll:
Could you please break these into individual PRs that fixes each bug separately?  Could you also provide on those PRs more context around what exactly is being fixed/why it was broken and the reasons you added the new features?

I agree suppressing the output of Echo seems counter-intuitive.  Is there no way to capture what you need without doing so?

egieseke (approved):
looks good.

egieseke (approved):
looks good.

onetechnical:
@btoll I think this fell through the cracks. Can you give it a re-review? I think generally a separate PR for bug fixes and one for feature adds is a good idea, though in this case the PR isn't that large. @bricerisingalgorand I'd prefer if the summary broke it out better into bug fixes/features as well.

onetechnical (commented):
I like some of the cleanups, but the quiet change is an interesting one. I can't say it _hurts_ anything to have, it just feels non-intuitive. New docker features are yay though.

btoll:
I really wish it would be broken into separate PRs, especially the new class.

It's hard to review some changes like this b/c there's no context around why something broke and how it was fixed.

As for the the quiet param, at the very least it needs a code in the comment that would help to explain why it's there.  It's really a head-scratcher w/o any context.

onetechnical:
> As for the the quiet param, at the very least it needs a code in the comment that would help to explain why it's there. It's really a head-scratcher w/o any context.

+1

PR #31, opened 2020-06-01
Add docker-compose to mule Dockerfile

bricerisingalgorand:
This is a useful addition and it's required by the algorand sdk pipelines


onetechnical (approved):
I'm actually surprised we haven't needed it yet.

btoll:
Should it go in the root of the project?

egieseke (approved):

PR #34, opened 2020-06-18
Add support for parameterized go version

egieseke:
Updated mule to support a parametrized go version retrieved from the go.mod file.  The go version is now included in the docker image tag.

### DIFF
--- a/mule/task/shell/__init__.py
+++ b/mule/task/shell/__init__.py
@@ -10,9 +10,10 @@ class IShellTask(ITask):
         if self.save_logs:
             print(f"running sub process {self.command}")
             command_logs = subprocess.run(self.command, capture_output=True, check=True)
-            print(f"command logs { command_logs.stdout.decode('utf-8')  }")
             return {
-                'logs': command_logs.stdout.decode('utf-8').rstrip("\n")
+                'stdout': command_logs.stdout.decode('utf-8').rstrip("\n"),
+                'stderr': command_logs.stderr.decode('utf-8').rstrip("\n"),
+                'returncode': command_logs.returncode
             }
         else:
             subprocess.run(self.command, check=True)
--- a/mule/task/shell/__init__.py
+++ b/mule/task/shell/__init__.py
@@ -3,9 +3,19 @@ import subprocess
 
 class IShellTask(ITask):
 
+    save_logs = False
+
     def execute(self, job_context):
         super().execute(job_context)
-        subprocess.run(self.command, check=True)
+        if self.save_logs:
+            print(f"running sub process {self.command}")
+            command_logs = subprocess.run(self.command, capture_output=True, check=True)
+            print(f"command logs { command_logs.stdout.decode('utf-8')  }")
+            return {
+                'logs': command_logs.stdout.decode('utf-8').rstrip("\n")
+            }
+        else:
+            subprocess.run(self.command, check=True)
 
 class Shell(IShellTask):
 
@@ -18,9 +28,11 @@ class Shell(IShellTask):
         self.command = args['command']
         if type(self.command) == str:
             self.command = self.command.split(' ')
+        if 'saveLogs' in args:
+            self.save_logs = args['saveLogs']
 
     def execute(self, job_context):
-        super().execute(job_context)
+        return super().execute(job_context)
 
 class Make(IShellTask):
 
--- a/mule/task/shell/docker.py
+++ b/mule/task/shell/docker.py
@@ -30,10 +30,10 @@ class Ensure(ITask):
         'version',
         'image',
         'dockerFilePath',
-        'arch',
-        'goVersion'
+        'arch'
     ]
-    buildContextPath= "."
+    buildContextPath = "."
+    goVersion = ""
 
     def __init__(self, args):
         super().__init__(args)
@@ -41,9 +41,10 @@ class Ensure(ITask):
         self.image = args['image']
         self.version = args['version']
         self.dockerFilePath = args['dockerFilePath']
-        self.goVersion = args['goVersion']
         if 'buildContextPath' in args:
             self.buildContextPath = args['buildContextPath']
+        if 'goVersion' in args:
+            self.goVersion = args['goVersion']
 
     def execute(self, job_context):
         super().execute(job_context)
--- a/mule/task/shell/docker.py
+++ b/mule/task/shell/docker.py
@@ -31,6 +31,7 @@ class Ensure(ITask):
         'image',
         'dockerFilePath',
         'arch',
+        'goVersion'
     ]
     buildContextPath= "."
 
@@ -40,10 +41,11 @@ class Ensure(ITask):
         self.image = args['image']
         self.version = args['version']
         self.dockerFilePath = args['dockerFilePath']
+        self.goVersion = args['goVersion']
         if 'buildContextPath' in args:
             self.buildContextPath = args['buildContextPath']
 
     def execute(self, job_context):
         super().execute(job_context)
         print('Checking to see if docker image is available')
-        docker.ensure(f"{self.image}:{self.version}", self.arch, self.buildContextPath, self.dockerFilePath)
+        docker.ensure(f"{self.image}:{self.version}", self.arch, self.goVersion, self.buildContextPath, self.dockerFilePath)
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,9 @@
+# go_algorand_ci code owners file
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+*       @btoll
+*       @bricerisignalgorand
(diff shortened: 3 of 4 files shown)


mule/task/shell/docker.py
```diff
@@ -31,6 +31,7 @@ class Ensure(ITask):
         'image',
         'dockerFilePath',
         'arch',
+        'goVersion'
```
btoll: I would rather have the version default to something reasonable rather than enforcing it as a param.

mule/task/shell/__init__.py
```diff
@@ -18,9 +28,11 @@ def __init__(self, args):
         self.command = args['command']
         if type(self.command) == str:
             self.command = self.command.split(' ')
+        if 'saveLogs' in args:
+            self.save_logs = args['saveLogs']
 
     def execute(self, job_context):
-        super().execute(job_context)
+        return super().execute(job_context)
```
btoll, line 36: Should this code be in another commit?  It seems out of scope.

btoll: That question is for all the new code in that file.

mule/task/shell/docker.py
egieseke: We want to make sure that it is specified for now.

mule/task/shell/__init__.py
egieseke: These changes are required to return the output of the get_golang_version.sh script.

mule/task/shell/docker.py
btoll: What if myself or another developer has a Dockerfile that doesn't care about go?

mule/task/shell/__init__.py
btoll: But what about the new `save_logs` option?

mule/task/shell/docker.py
btoll: This is also a breaking change.  All the yaml files will need to be updated, which means anyone using this will find that their process is broken.

mule/task/shell/__init__.py
egieseke: That is also required to fetch the go version.

btoll: It would be nice to have an example in the PR description that shows how this is used.  I'm assuming from your draft PR here https://github.com/algorand/go-algorand/pull/1185 that this is what it would look like:
```
  - task: shell.docker.Ensure
    name: alpine
    image: algorand/go-algorand-ci-linux
    version: '{{ docker.Version.outputs.version }}-{{ shell.Shell.go-version.outputs.logs }}'
    goVersion: '{{ shell.Shell.go-version.outputs.logs }}'
    dockerFilePath: docker/build/cicd.alpine.Dockerfile
    dependencies: shell.Shell.go-version docker.Version
```
My thoughts on that are that:
- It would be great to not have that as a required param.
- It would be simpler just to specify the version as `goVersion: 1.14` or whatever.
- It seems that the `save_logs` option is just for accessing the go version, not as a general "get all my logs" as the name implies.  If this were to be kept, would it be better to name it `{{ shell.Shell.go-version.outputs.goVersion }`?

bricerisingalgorand: The reason a script is setting the go version is because the travis pipeline needs to determine which go version to use as well. Also, using this script we can determine what version of go to use from the respective field in the go.mod file in order to test on the minimum version we're supporting

bricerisingalgorand: The reason we want to use a script is so that we can get the go version from the go.mod file that declares the lowest version our project is compatible with. Also, the travis job needs to know the go version as well, so using a script both travis and mule can get the go version from the same place

mule/task/shell/docker.py
egieseke: @btoll  OK, made goVersion optional.  Good thinking.

btoll: Great, thanks.

bricerisingalgorand (dismissed):

mule/task/shell/__init__.py
btoll: It'd be good to have the output be more general.  How about we capture all the data and have it as an object:
```
shell.Shell.go-version.outputs.logs {}
```

onetechnical: I think that would imply an opinionated output from an arbitrary shell command. I think I'd prefer output or something, as "logs" have other connotations. (This makes for an ugly outputs.output, but I can't think of a better option.)

onetechnical (dismissed):
I approve of this, with a caveat. The goVersion handling here is tightly coupled to docker implementation, which makes it less of a general use case and more of a specific use case. We should refactor to be more independent, because I don't like forcing this to be go-specific. Can you create a ticket for that?

mule/task/shell/__init__.py
btoll: My thought is that it would be ok to have all of the params that were passed available if needed or wanted in a `logs` object, and this could simply access what it needs via `logs.goVersion`.

As your comment below mentions, the fixes and features that I'm seeing are addressing specific use cases that we may need internally, but we really need to be thinking about this more as a framework.

btoll:
This ticket should be linked with this one, as it's using the goVersion log that's implemented here: https://github.com/algorand/go-algorand/pull/1185/

btoll (changes requested):
I'm going to block this for a couple of reasons.
1.  I had raised a question the end of last week about how this is returning its "log" information, but I don't feel that it's been sufficiently addressed.  I understand that the current approach works, but I don't feel that it's sufficient.
2.  The other PR that I linked to is relying syntactically on this approach, so if we don't fix this now it will create more churn than just in the `mule` repo.

mule/task/shell/__init__.py
bricerisingalgorand: I don't think I understand. What data are you looking for from a shell command other than the logs?

As far as task inputs go, those are available for every task already using `task_id.inputs.field_name`

btoll: My presumption was that this added the ability to pass any number of commands that would then be available in an object (I thought `logs` was that object).  Now I also understand John's comment as well.

I must admit I'm not crazy about this approach.  It seems bolted on just to make this use case work.

If you want more data from commands to be available to subsequent processes, currently you'll have to write them one after the other, which further increases the size of these yaml files.  Also, it increases the number of tasks that need to be added to another one's dependency config (and then the further re-running of those tasks).

bricerisingalgorand: But what more data are you looking for?

It's ok that you want more features out of the `shell.Shell` task, but right now all we need is a way to use logs from scripts that task has executed. If we want that task to support executing multiple shell commands, we can either make a story for implementing that feature or work on a story that requires it.

btoll: I'm saying the opposite, actually.  I don't think adding more functionality to the `shell.Shell` task is a good idea, it's just a way of quickly getting this to work.

If this is to be refactored to also return the output of the command, I would think the next question to be asked is what if I need to capture the output of multiple commands?  That would cut down on the size of the yaml file, which is something I hope we're all concerned about.

Since this is new functionality, I think it's reasonable to ask that question now and possibly accomodate that approach rather than kicking that can down the line.

In the meantime, I think a simple improvement would be the naming.  For example, call it `stdout` instead of `logs`, which is more descriptive of what is being captured.  So, the output could be referenced like:
```
shell.Shell.go-version.outputs.stdout
```
Also, maybe just always return it, that way it saves the user from having to explicit specify that they want it.

onetechnical: I like the idea of using stdout instead of logs. I'm not sure I understand the multiple command bit; would we pass it an array of arrays of commands or something?

bricerisingalgorand: Supporting always printing script logs and storing them would be difficult because the process outputs are appended to IO streams. This means that in order to do that we would need to manage our own stream and handle frequent forwarding to stdout ourselves. We don't have a use case here where that is necessary, so that's another thing we should probably make another story if we actually need that. We can go ahead and rename the field to stdout though, and include stderr while we're at it

egieseke: Split return of IShellTask.execute into 3 parts for `stdout`, `stderr`, and `returncode`.  All can be accessed by the caller.

btoll: Sounds good, thanks guys.

btoll (approved):

onetechnical (approved):
I like the change!

btoll:
@onetechnical Yes, multiple commands could look something like this:
```
  - task: shell.Shell
    name: foo
    stdout: True
    command:
      -./get_golang_version.sh
      -./foo.sh
      -./bar.sh
```
rather than:
```
  - task: shell.Shell
    name: golang
    stdout: True
    command: ./get_golang_version.sh

  - task: shell.Shell
    name: foo
    stdout: True
    command: ./foo.sh

  - task: shell.Shell
    name: bar
    stdout: True
    command: ./bar.sh
```

PR #35, opened 2020-06-24
A mule task for starting indexer nodes

bricerisingalgorand:
This is a mule task that allows us to start indexer instances


mule/util/algorand_util.py
```diff
@@ -147,3 +147,52 @@ def goal(data_dir, kmd_dir, args, bin_dir=None):
...
+            '--pidfile', pid_file
+        ])
+
+    algorand_indexer(algorand_indexer_args, bin_dir, log_file_name)
+
+def build_indexer_postgress_connection_string(postgres):
+    postgress_connection_string = []
+    for field in postgres.items():
+        postgress_connection_string.append(f"{field[0]}={field[1]}")
+    return ' '.join(postgress_connection_string)
```
btoll, line 198: These functions are so particular to a class that I think it would be better to have them in the class.

btoll:
Overall I think it's fine, but if we're keeping the idea of making this a general purpose tool I'm not sure I'd approach it like this.  I think it would be better to pull out the algorand specific code into its own location and then simplly pass an object into `Start` that you need to start.

onetechnical (approved):
Overall the approach here looks fine to me, but I'm presuming you've tested it as well.

bricerisingalgorand:
> Overall I think it's fine, but if we're keeping the idea of making this a general purpose tool I'm not sure I'd approach it like this. I think it would be better to pull out the algorand specific code into its own location and then simplly pass an object into `Start` that you need to start.

We've talked about breaking algorand code into a separate algorand-tasks repo, but since we're the only ones currently using mule I think it makes sense to keep them together for now. If we want we can have a story to break them apart. Mule supports outside task definitions pretty conveniently

btoll:
> > Overall I think it's fine, but if we're keeping the idea of making this a general purpose tool I'm not sure I'd approach it like this. I think it would be better to pull out the algorand specific code into its own location and then simplly pass an object into `Start` that you need to start.
> 
> We've talked about breaking algorand code into a separate algorand-tasks repo, but since we're the only ones currently using mule I think it makes sense to keep them together for now. If we want we can have a story to break them apart. Mule supports outside task definitions pretty conveniently

Sounds good.

mule/util/algorand_util.py
bricerisingalgorand: I'm not totally sure what you mean. Are you saying these could go into the algorand.indexer.Start task? This function is going to be required later on by the indexer import features. I think it's ideal though to keep the responsibility of the task objects as slim as possible to keep the code more readable

egieseke (approved):
Looks good.

mule/util/algorand_util.py
btoll: No, they should be in the class itself.

PR #37, opened 2020-07-02
Promote docker to first class object

btoll:
Broad overview:
- new `Docker` class
- new top-level `agents` definition
  - `agents` can accept a `buildArgs` list
- `docker.Shell` task can pass in an `agent` config that is name of an existing docker image (`name`:`version`)

---

This PR has a new `Docker` class which has subsumed the `Ensure` and `Version` classes.  These latter classes don't need to be exposed as they are plumbing operations that can be managed by the new `Docker` class.  Note that the underlying functionality hasn't changed, only what needs to be exposed.

`docker.Shell` and `docker.Make` are now child classes of the new `Docker` class, which does all the heavy lifting by abstracting away the aforementioned operations and others.  The main functional difference is that the `Docker` class is not following the path that most classes go down, which is to inspect their dependencies and have the `mule` framework then order them and then call their constructors and `execute` methods with the appropriate task configs.

This isn't necessary for a couple of reasons.  One, there shouldn't be any configs that need to be rendered by `pystache`.  This is because the new first-class `Docker` objects will already contain everything it needs to bootstrap itself.  Two, there shouldn't be any more dependencies that the `Docker` object needs.  This can be subject to change, of course, but I currently don't believe we have any use cases that need dependencies other than `Version` and `Ensure`, and I don't believe there is a current need for anything other than those.

In addition, the `agent` can accept a `buildArgs` config that will be evaluated and passed through to the subprocess that calls `docker build`.  See below for an example.

This PR also lifts the functions that the `Docker` class needs out of the utility function and into the class definition itself.

Example:
```
agents:
  - name: ubuntu
    dockerFilePath: Dockerfile.test
    image: algorand/go-algorand-ci-linux-ubuntu
    version: README.md
    buildArgs:
      - GOLANG_VERSION=`./get_golang_version.sh`
      - PWD=`pwd`
      - Z=ZED
    env:
      - A=${A}
      - B=${B}
    workDir: /projects/go-algorand

  - name: centos
    dockerFilePath: Dockerfile.test
    image: algorand/go-algorand-ci-linux-centos
    version: Makefile
    workDir: /projects/go-algorand

tasks:
  - task: docker.Shell
    command: cat /etc/*release
    agent: amd64/alpine:latest

  - task: docker.Make
    name: build
    target: hello
    agent: ubuntu

jobs:
  cat-release:
    tasks:
    - docker.Shell

  hello:
    tasks:
      - docker.Make.build
```


mule/error/messages.py
```diff
@@ -1,20 +1,22 @@
...
 FIELD_NOT_FOUND="Field {} was not found"
-FIELD_VALUE_WRONG_TYPE="Field {} is the wrong type, expected {} got {}"
 FIELD_VALUE_COULD_NOT_BE_VALIDATED="Value of field {} could not be validated:\n{}"
+FIELD_VALUE_WRONG_TYPE="Field {} is the wrong type, expected {} got {}"
 JOB_CONFIG_WRONG_TYPE="Job {} is the wrong type, expected {} got {}"
 JOB_MISSING_REQUIRED_FIELDS="Job {} is missing required fields.\nThe following fields are required {}"
-TASK_FIELD_IS_WRONG_TYPE="Task {} has a field '{}' with the wrong type, expected {} got {}"
+JOB_NOT_FOUND="Could not find job for name {}"
+MACHINE_DUPLICATE_NAME="Machine name '{}' not unique"
+MACHINE_MISSING_NAME="Could not find machine name '{}'"
```
btoll, line 14: Only added the above two messages (`MACHINE_*`), the rest of the noise is b/c the messages have been sorted for better readability.

bricerisingalgorand:
I think it would be advantageous for us to remove the concept of IDockerTask and IShellTask, and instead depend on an Agent interface with a run command that can accept shell strings. Then tasks like Make can instantiate objects implementing the IAgent interface based on the agent params in the job_context (with ShellAgent being the default). The Agent objects would be similar to the Machine class you wrote for docker agents, but calling them Agents instead would be more consistent with the yaml language. This way we won't have to add a new tasks for each of the agent type we support. What do you think?

Nice work doing this so quickly by the way!

btoll:
> I think it would be advantageous for us to remove the concept of IDockerTask and IShellTask, and instead depend on an Agent interface with a run command that can accept shell strings. Then tasks like Make can instantiate objects implementing the IAgent interface based on the agent params in the job_context (with ShellAgent being the default). The Agent objects would be similar to the Machine class you wrote for docker agents, but calling them Agents instead would be more consistent with the yaml language. This way we won't have to add a new tasks for each of the agent type we support. What do you think?
> 
> Nice work doing this so quickly by the way!

Yes, sounds good, I'll take a look at that.

mule/task/docker/__init__.py
```diff
@@ -1,39 +1,138 @@
...
-        ('docker.shell', str),
+class Docker(ITask):
+    required_fields = [
+        'image',
     ]
 
-    command = ''
+    machine = {
+        'arch': 'amd64',
+        'buildArgs': [],
```
btoll, line 20: Allowing for `buildArgs` avoids the necessity to create hard-coded params like `goVersion` and also leverages Docker's own API.

mule/task/docker/__init__.py
```diff
@@ -1,39 +1,138 @@
...
+    def evaluateBuildArgs(self):
+        args = []
+        for build_arg in self.machine['buildArgs']:
+            symbol, value = build_arg.split('=')
+            if value[0] == '`':
+                ret = subprocess.run(value.strip('`'), capture_output=True)
+                args.append(symbol + '=' + ret.stdout.decode('utf-8').rstrip('\n'))
+            else:
+                args.append(build_arg)
+        return args
```
btoll, line 105: New method to allow for unlimited number of build args to be passed.  I chose to use backticks (`) because of their well-known history for evaluation in Unix environments.

```
agents:
  - name: ubuntu
    dockerFilePath: Dockerfile.test
    image: algorand/go-algorand-ci-linux-ubuntu
    version: README.md
    buildArgs:
      - GOLANG_VERSION=`./get_golang_version.sh`
      - PWD=`pwd`
      - Z=ZED
    env:
      - A=${A}
      - B=${B}
    workDir: /projects/go-algorand
```

btoll:
@bricerisingalgorand I didn't do all that you suggested for a couple of reasons.  One, I wanted, at least for now, to keep the different `Make` classes in their own namespaces.  Two, I didn't want this to get much larger than it currently it is.  This does all that I wanted a first pass to do, and we can always do more.

In addition, I added the `buildArgs` config because I needed to address the functionality of the `goVersion` param that was recently added which was lost with the erasure of the `Version` class..  I debated doing this in another PR, but I didn't want to create a dependency.

bricerisingalgorand:
> @bricerisingalgorand I didn't do all that you suggested for a couple of reasons. One, I wanted, at least for now, to keep the different `Make` classes in their own namespaces. Two, I didn't want this to get much larger than it currently it is. This does all that I wanted a first pass to do, and we can always do more.
> 
> In addition, I added the `buildArgs` config because I needed to address the functionality of the `goVersion` param that was recently added which was lost with the erasure of the `Version` class.. I debated doing this in another PR, but I didn't want to create a dependency.

Sounds good. This PR is a great first step, and we can always decide to do that decoupling later on if we want to. Just eyeballing it the code looks good. Could you also update the yamls in `test/test-yamls` to match this implementation? Once we know those test cases still work I think we'll be ready to merge 👍

mule/task/docker/__init__.py
bricerisingalgorand: This is awesome, it never crossed my mind to see if command substitution would work!

bricerisingalgorand: This is awesome, it never crossed my mind to support command substitution!

btoll:
> > @bricerisingalgorand I didn't do all that you suggested for a couple of reasons. One, I wanted, at least for now, to keep the different `Make` classes in their own namespaces. Two, I didn't want this to get much larger than it currently it is. This does all that I wanted a first pass to do, and we can always do more.
> > In addition, I added the `buildArgs` config because I needed to address the functionality of the `goVersion` param that was recently added which was lost with the erasure of the `Version` class.. I debated doing this in another PR, but I didn't want to create a dependency.
> 
> Sounds good. This PR is a great first step, and we can always decide to do that decoupling later on if we want to. Just eyeballing it the code looks good. Could you also update the yamls in `test/test-yamls` to match this implementation? Once we know those test cases still work I think we'll be ready to merge

Thanks for the reminder, I'll update those.

btoll:
The `stash.yaml` test failed.  I didn't want to include it as part of this ticket since the failures are unrelated.  I created a ticket for it here: https://github.com/algorand/go-algorand-internal/issues/653

I wasn't sure how to setup to test `algorand-test.yaml`.  The failures are all missing dirs and are unrelated to my changes.  Also, I removed `tasks-parameterization-test.yaml` since it is testing the return value of `Version` which is now obsolete.

Everything else tested fine.

mule/task/__init__.py
```diff
@@ -85,7 +84,10 @@ class Job(ITask):
     ]
 
     required_typed_fields = [('tasks', list)]
-    optional_typed_fields = [('configs', dict)]
+    optional_typed_fields = [
+        ('agent_configs', list),
+        ('configs', dict)
```
egieseke, line 89: Should `configs` be renamed to something more descriptive?

btoll: Those refer to the jobs configs that are optionally defined in the yaml file, such as:
```
jobs:
  package:
    configs:
      arch: amd64
    tasks:
      - shell.docker.Ensure.rpm
      - docker.Make.rpm
```
I think it's probably fine to keep that name.  What did you have in mind?

egieseke: maybe jobs_configs

btoll: Oh, I see, sure, I can change that.

egieseke (approved):
Assuming that Brice's questions have been resolved, looks good.

bricerisingalgorand (approved):

PR #39, opened 2020-08-10
Updated mule docs to include new `agents` block

btoll:


egieseke (dismissed):
Looks good.

egieseke (dismissed):

onetechnical (dismissed):

docs/agents/README.md
```diff
@@ -0,0 +1,34 @@
+# Agents
+
+## Description
+
+This task executes a make target from within a docker container. This task requires that there is a Makefile in the directory from where the task was called.
```
bricerisingalgorand: Should this be something like "This is used to define an agent that can execute mule tasks" instead?

btoll: Yeah, you're right, will change.

egieseke (approved):

bricerisingalgorand (approved):

PR #41, opened 2020-08-11
Automate pushing new version to pypi and docker hub

btoll:


scripts/push.sh
```diff
@@ -0,0 +1,24 @@
...
+
+VERSION=$1
+
+if [[ -z $VERSION ]]
+then
+    VERSION=${1:-$(sed -n 's/.*=\s*''\([0-9.]*\)''/\1/p' ./mule/__init__.py)}
+    # Strip single quote from beginning.
+    VERSION="${VERSION#\'}"
+    # Strip single quote from end.
+    VERSION="${VERSION%\'}"
```
egieseke, line 13: I wonder if awk would work better for this.

egieseke (dismissed):
looks good.

scripts/push.sh
btoll: I looked at it, but it's actually more complicated.  It involves a multi-line awk function.

scripts/push.sh
```diff
@@ -0,0 +1,24 @@
...
+fi
+
+# Push to pypi.
+python3 setup.py bdist_wheel
+twine upload "dist/mulecli-$VERSION-py3-none-any.whl"
+
+# Push to docker hub for pipelines.
```
bricerisingalgorand, line 23: This looks great! We may need to add a sleep here though (maybe `sleep 10`). I've noticed occasionally the pypi repo takes a few seconds to update after the file is uploaded, so the docker build may have issues if that happens

btoll: Ok, I'll update.

bricerisingalgorand (approved):

egieseke (approved):
looks good.

PR #42, opened 2020-10-07
Fix docker ensure

bricerisingalgorand:
Workdir was being used for the docker build context path. The workDir field is used to say what directory on the created container a mule task should be on, rather than what directory docker should build an image from. I added a context field where this setting can be added in the future.

### DIFF
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -25,6 +25,7 @@ class Docker(ITask):
         'version': '',
         'volumes': [],
         'workDir': '/project',
+        'context': '.'
     }
 
     def __init__(self, args):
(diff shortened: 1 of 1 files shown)


btoll (approved):

onetechnical (approved):

mule/task/docker/__init__.py
```diff
@@ -39,7 +40,7 @@ def build(self, image):
             build_args_str = f"--build-arg {' --build-arg '.join(build_args)}"
         if tags is not None and len(tags) > 0 :
             tags_str = f"-t {' -t '.join(tags)}"
-        docker_command = f"docker build {build_args_str} {tags_str} -f {self.machine['dockerFilePath']} {self.machine['workDir']}"
+        docker_command = f"docker build {build_args_str} {tags_str} -f {self.machine['dockerFilePath']} {self.machine['context']}"
```
onetechnical, line 43: Did you want to add support for workdir here?

onetechnical: nvm, I don't think this applies in a build context.

PR #44, opened 2020-10-07
Brice/support multiple yamls

bricerisingalgorand:
With this we'll be able to support multiple mule yamls and deep merge them in the order that they were passed in:

for example:
```
mule -f test/test-yamls/valid/overrides-test.yaml -f test/test-yamls/valid/overrides-test.yaml echo-configs-override
```

### DIFF
--- a/mule/driver.py
+++ b/mule/driver.py
@@ -2,7 +2,7 @@ import yaml
 import mule.parser
 import mule.validator as validator
 import mule.util.yaml.env_var_loader as yaml_util
-from mule.util import JobContext, file_util
+from mule.util import JobContext, file_util, update_dict
 from mule.error import messages
 from mule.task import Job
 from termcolor import cprint
@@ -11,7 +11,10 @@ import sys
 def main():
     args = mule.parser.parseArgs()
     try:
-        mule_config = file_util.readYamlFile(args.file)
+        mule_yamls = args.file
+        if len(args.file) == 0:
+            mule_yamls.append('mule.yaml')
+        mule_config = _read_mule_yamls(mule_yamls)
         parsed_mule_config = yaml_util.readYamlWithEnvVars(yaml.dump(mule_config))
         jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
         if args.list_jobs:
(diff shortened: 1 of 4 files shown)


mule/driver.py
```diff
@@ -29,6 +32,16 @@ def main():
...
 
+def _read_mule_yamls(mule_yamls):
+    mule_config = {}
+    for mule_yaml in mule_yamls:
+        update_dict(
+            mule_config,
+            file_util.readYamlFile(mule_yaml),
+            overwrite_lists=False
+        )
+    return mule_config
```
egieseke, line 43: Is it possible that there could be conflicts in the mule.yaml files?  Should there be a warning if conflicts are found?

egieseke (approved):
Looks good.

mule/parser.py
```diff
@@ -16,7 +16,8 @@ def parseArgs():
     parser.add_argument(
         '-f',
         '--file',
-        default = 'mule.yaml',
+        action='append',
+        default = [],
```
btoll, line 20: What about:
```
default = ['mule.yaml']
```

mule/driver.py
bricerisingalgorand: It's implemented such that subsequent files overwrite values in previous files, except when both values are lists the contents are appended

mule/parser.py
bricerisingalgorand: If we have this here as default, it'll be included in the list when files are passed in. I'm not sure that's what people expect when they provide one or more yaml files to the cli

for example, this command would produce this list
```
mule -f test/mule.yaml test

[ "mule.yaml", "test/mule.yaml" ]
```

mule/driver.py
```diff
@@ -11,7 +11,10 @@
 def main():
     args = mule.parser.parseArgs()
     try:
-        mule_config = file_util.readYamlFile(args.file)
+        mule_yamls = args.file
+        if len(args.file) == 0:
+            mule_yamls.append('mule.yaml')
```
btoll, line 16: Could be avoided if setting the default like I mentioned below.

mule/parser.py
btoll: Makes sense.

mule/driver.py
btoll: This is moot in light of what you wrote to first comment.

btoll (approved):

onetechnical (approved):

PR #45, opened 2020-10-30
Update algorand install to support nightly channel

bricerisingalgorand:
The installer needs to select the correct genesis file when installing the nightly channel


egieseke (approved):
Looks good.

onetechnical (approved):

PR #47, opened 2020-11-01
Added --list-agents function and verbose flag

btoll:
Added a new `_list_agents` function to keep parity with listing jobs and tasks.  In addition, it adds a `--verbose` flag which allows for seeing more detail about all `list*` functions.

`mule -f package.yaml --list-agents`
`mule -f package.yaml --list-agents --verbose`

### DIFF
--- a/mule/driver.py
+++ b/mule/driver.py
@@ -1,12 +1,13 @@
-import yaml
+from termcolor import cprint
+import sys
+
 import mule.parser
 import mule.validator as validator
 import mule.util.yaml.env_var_loader as yaml_util
 from mule.util import JobContext, file_util, update_dict
+from mule.util.algorand_util import pjson
 from mule.error import messages
 from mule.task import Job
-from termcolor import cprint
-import sys
 
 def main():
     args = mule.parser.parseArgs()
@@ -15,14 +16,19 @@ def main():
         if len(args.file) == 0:
             mule_yamls.append('mule.yaml')
         mule_config = _read_mule_yamls(mule_yamls)
-        parsed_mule_config = yaml_util.readYamlWithEnvVars(yaml.dump(mule_config))
-        jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
-        if args.list_jobs:
-            _list_jobs(jobs_config)
-        elif args.list_tasks:
-            _list_tasks(task_configs)
+
+        if args.list_agents:
+            parsed_mule_config = yaml_util.readYaml(mule_config)
+            _list_agents(parsed_mule_config["agents"], args.verbose)
         else:
-            _execute_job(jobs_config, task_configs, agent_configs, args.JOB)
+            parsed_mule_config = yaml_util.readYaml(mule_config, raw=False)
+            jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
+            if args.list_jobs:
+                _list_jobs(jobs_config, args.verbose)
+            elif args.list_tasks:
+                _list_tasks(task_configs, args.verbose)
+            else:
+                _execute_job(jobs_config, task_configs, agent_configs, args.JOB)
     except Exception as error:
         cprint(
             messages.MULE_DRIVER_EXCEPTION.format(args.JOB, args.file, error),
@@ -42,16 +48,29 @@ def _read_mule_yamls(mule_yamls):
         )
     return mule_config
 
-def _list_jobs(jobs_config):
-    for job in jobs_config.keys():
-        print(job)
-
-def _list_tasks(task_configs):
-    for task in task_configs:
-        if 'name' in task.keys():
-            print(f"{task['task']}.{task['name']}")
+def _list_agents(agent_configs, verbose):
+    for agent in agent_configs:
+        if verbose:
+            print(agent['name'], pjson(agent_configs))
         else:
-            print(task['task'])
+            print(agent['name'])
+
+def _list_jobs(jobs_config, verbose):
+    for job in jobs_config.keys():
+        if verbose:
+            print(job, pjson(jobs_config[job]))
+        else:
+            print(job)
+
+def _list_tasks(task_configs, verbose):
+    if verbose:
+        print(pjson(task_configs))
+    else:
+        for task in task_configs:
+            if 'name' in task.keys():
+                print(f"{task['task']}.{task['name']}")
+            else:
+                print(task['task'])
 
 def _execute_job(jobs_config, task_configs, agent_configs, job):
     if not job in jobs_config:
--- a/mule/driver.py
+++ b/mule/driver.py
@@ -4,8 +4,7 @@ import sys
 import mule.parser
 import mule.validator as validator
 import mule.util.yaml.env_var_loader as yaml_util
-from mule.util import JobContext, file_util, update_dict
-from mule.util.algorand_util import prettify_json
+from mule.util import JobContext, file_util, prettify_json, update_dict
 from mule.error import messages
 from mule.task import Job
 
--- a/mule/driver.py
+++ b/mule/driver.py
@@ -13,22 +13,19 @@ def main():
     args = mule.parser.parseArgs()
     try:
         mule_yamls = args.file
-        if len(args.file) == 0:
+        if len(mule_yamls) == 0:
             mule_yamls.append('mule.yaml')
         mule_config = _read_mule_yamls(mule_yamls)
-
+        parsed_mule_config = yaml_util.read_yaml(mule_config, raw=yaml_read_raw(args))
+        agent_configs, jobs_config, task_configs = validator.get_validated_mule_yaml(parsed_mule_config)
         if args.list_agents:
-            parsed_mule_config = yaml_util.read_yaml(mule_config)
-            _list_agents(parsed_mule_config["agents"], args.verbose)
+            _list_agents(agent_configs, args.verbose)
+        elif args.list_jobs:
+            _list_jobs(jobs_config, args.verbose)
+        elif args.list_tasks:
+            _list_tasks(task_configs, args.verbose)
         else:
-            parsed_mule_config = yaml_util.read_yaml(mule_config, raw=False)
-            jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
-            if args.list_jobs:
-                _list_jobs(jobs_config, args.verbose)
-            elif args.list_tasks:
-                _list_tasks(task_configs, args.verbose)
-            else:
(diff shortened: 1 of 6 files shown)


mule/driver.py
```diff
@@ -42,16 +48,29 @@ def _read_mule_yamls(mule_yamls):
...
+
+def _list_tasks(task_configs, verbose):
+    if verbose:
+        print(pjson(task_configs))
+    else:
+        for task in task_configs:
+            if 'name' in task.keys():
+                print(f"{task['task']}.{task['name']}")
+            else:
+                print(task['task'])
```
btoll, line 69: The noise here is simply ordering the functions:
- _list_agents
- _list_jobs
- _list_tasks

mule/util/algorand_util.py
```diff
@@ -196,3 +197,7 @@ def build_indexer_postgress_connection_string(postgres):
     for field in postgres.items():
         postgress_connection_string.append(f"{field[0]}={field[1]}")
     return ' '.join(postgress_connection_string)
+
+def pjson(raw_json):
+    return json.dumps(raw_json, indent=4, sort_keys=False)
```
btoll: Prettify json helper.

mule/util/yaml/env_var_loader.py
```diff
@@ -18,12 +19,17 @@ def path_constructor(loader, node):
...
-def readYamlWithEnvVars(stream):
-    loader = getYamlLoaderWithEnvVars()
+def readYaml(mule_config, raw=True):
+    loader = yaml
+    stream = yaml.dump(mule_config)
+
+    if not raw:
+        loader = resolveEnvVars()
+
     return loader.load(stream, Loader=yaml.FullLoader)
```
btoll: Since we're now interested in reading the yaml for both raw (doesn't evaluate env vars, for instance for _list_agents) and not raw (evaluate env vars, the current behavior), we can simplify the file by just calling a common `readYaml` function and telling it which behavior we want.

This also allows us to remove the `yaml` import from `driver.py` and do all yaml evaluations in a single file.

mule/driver.py
```diff
@@ -15,14 +16,19 @@ def main():
...
         else:
-            _execute_job(jobs_config, task_configs, agent_configs, args.JOB)
+            parsed_mule_config = yaml_util.readYaml(mule_config, raw=False)
+            jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
+            if args.list_jobs:
+                _list_jobs(jobs_config, args.verbose)
+            elif args.list_tasks:
+                _list_tasks(task_configs, args.verbose)
+            else:
+                _execute_job(jobs_config, task_configs, agent_configs, args.JOB)
```
btoll: Since the `list_agents` operations doesn't require evaluating any env vars, we should tell `readYaml` not to do it (`readYaml` defaults to NOT evaluating the env vars).

Since the other operations either don't care OR do need the env vars to be evaluated, put them in the `else` condition.

mule/util/algorand_util.py
```diff
@@ -196,3 +197,7 @@ def build_indexer_postgress_connection_string(postgres):
     for field in postgres.items():
         postgress_connection_string.append(f"{field[0]}={field[1]}")
     return ' '.join(postgress_connection_string)
+
+def pjson(raw_json):
```
egieseke: consider a less cryptic function name, for example, pretty_print_json.

mule/util/yaml/env_var_loader.py
```diff
@@ -18,12 +19,17 @@ def path_constructor(loader, node):
         return ''
     return value
 
-def getYamlLoaderWithEnvVars():
+def resolveEnvVars():
```
egieseke: Since we are renaming this function, take the opportunity to make the naming consistent.  For example, resove_env_vars() and read_yaml().  https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

mule/util/algorand_util.py
btoll: Will change.

mule/util/yaml/env_var_loader.py
btoll: I like your thinking, will change :)

egieseke (dismissed):
Looks good.

mule/driver.py
```diff
@@ -15,14 +16,19 @@ def main():
...
             mule_yamls.append('mule.yaml')
         mule_config = _read_mule_yamls(mule_yamls)
-        parsed_mule_config = yaml_util.readYamlWithEnvVars(yaml.dump(mule_config))
-        jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
-        if args.list_jobs:
-            _list_jobs(jobs_config)
-        elif args.list_tasks:
-            _list_tasks(task_configs)
+
+        if args.list_agents:
```
bricerisingalgorand, line 20: We should probably validate the mule yaml before this right? Rather than doing an if else

mule/driver.py
bricerisingalgorand: Are we preferring that env vars don't get evaluated in the list agents option? I would that that would be expected behavior, since it would be useful to see what gets configured in a spun up agent

mule/util/algorand_util.py
bricerisingalgorand: This looks useful, but might be more approrpiate in `mule/util/__init__.py`, since this file is for algorand utilities

mule/util/yaml/env_var_loader.py
```diff
@@ -18,12 +19,17 @@ def path_constructor(loader, node):
         return ''
     return value
 
-def getYamlLoaderWithEnvVars():
+def evaluate_env_vars():
```
bricerisingalgorand: This function name doesn't quite make sense since what this returns is a yaml loader, rather than evaluating env vars on something

mule/driver.py
bricerisingalgorand: Maybe we should have a cli option that generally doesn't evaluate env vars? Then it could be used for any mule command

btoll: The idea here is not to evaluate potentially sensitive information such as aws creds unless desired.  This is a good idea when either screen sharing, pair programming or otherwise just wanting to prevent people from shoulder-surfing.

btoll: Giving it the `--verbose` flag will evaluate these env vars.  I don't know if you recall, but I went into detail on these behaviors on a standup call.

mule/util/algorand_util.py
btoll: Will move.

mule/util/yaml/env_var_loader.py
btoll: Will change the name.

mule/driver.py
btoll: The agents aren't being validated in the `getValidatedMuleYaml` function.  I probably should have added validation when I added the list-agents to the code in a previous PR.

I'll do all that now.

mule/driver.py
bricerisingalgorand: yea but what i'm thinking might work better for us than having the if agents don't evaluate, else evaluate, we could have an option like `--ignore-env` on the mule cli that sets a boolean and is passed to the `raw` arguement. Then we can use both `mule --list-agents` and `mule --list-agents --ignore-env`. That may be useful for other utilities as well like describing tasks/jobs later on

btoll: That may be ok, but for this ticket I want to keep the scope to the agents themselves.

Also, the need for this developed because I wanted a nice way to dump out the env vars needed for a particular job backed by a particular agent, and in particular I wanted to target the `env` block in the agent config.

I'm sure there are use cases for it, but in the mule yaml files I've been writing for packaging (and others), the env vars I need to know about are in the agents and aren't in either the jobs or the tasks.  For my cases, I just don't see a need to evaluate anything in those particular blocks, at least not currently.

bricerisingalgorand: Well, I think in the meantime we could go with something like this, to avoid the main method having to determine read settings. Also I think we should still validate the mule.yaml when reading listing agents so that we can verify that we're looking at a valid mule.yaml

```python3
def main():
    try:
        mule_yamls = args.file
        if len(args.file) == 0:
        mule_config = _read_mule_yamls(mule_yamls)
        parsed_mule_config = yaml_util.read_yaml(mule_config, raw=yaml_read_raw())
        jobs_config, task_configs, agent_configs = validator.getValidatedMuleYaml(parsed_mule_config)
        if args.list_agents:
            _list_agents(agent_configs, args.verbose)
        elif args.list_jobs:
            _list_jobs(jobs_config, args.verbose)
        elif args.list_tasks:
            _list_tasks(task_configs, args.verbose)
        else:
            _execute_job(jobs_config, task_configs, agent_configs, args.JOB)
    except Exception as error:
        cprint(
            messages.MULE_DRIVER_EXCEPTION.format(args.JOB, args.file, error),
            'red',
            attrs=['bold'],
            file = sys.stderr
        )
        raise error

def yaml_read_raw(args):
    return args.list_agents
```

btoll: yes, I'm working on validation, see my comment below.

btoll: I had separated out the list-agents since, when I currently created this ticket, I wasn't evaluating the env vars for the list-agents but was doing that in the subsequent ticket.  The idea at the time was to avoid the hit of evaluation if we didn't need it.

I like your approach above and will add it to the PR.

btoll: By the way, here is what it looks like when you evaluate for agents:
```
deb [
    {
        "buildArgs": [
            "GOLANG_VERSION=`./scripts/get_golang_version.sh`"
        ],
        "dockerFilePath": "docker/build/cicd.ubuntu.Dockerfile",
        "env": [
            "AWS_ACCESS_KEY_ID=derp",
            "AWS_SECRET_ACCESS_KEY=derp",
            "",
            "",
            "",
            ""
        ],
```
I have a way of doing it in my other PR that is quite nice.

mule/validator/__init__.py
```diff
@@ -22,46 +22,51 @@ def getValidatedMuleConfigFile():
...
+
+    mule_configs = [
+        ("jobs", jobs_configs, validate_block),
+        ("tasks", task_configs, validate_tasks)
+    ]
+
+    if len(agent_configs):
+        mule_configs.append(("agents", agent_configs, validate_list))
+
+    for (name, config, fn) in mule_configs:
```
bricerisingalgorand, line 47: oh this is pretty cool

bricerisingalgorand (approved):

mule/validator/__init__.py
btoll: :bananadance:

egieseke (approved):

PR #49, opened 2020-11-08
Added list-env functionality for printing out a job's customizable env vars

btoll:
```
$ mule -f package-sign.yaml --list-env package-sign
deb:
- AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
- BRANCH=$BRANCH
- CHANNEL=$CHANNEL
- S3_SOURCE=$S3_SOURCE
- VERSION=$VERSION
```

Adding the `--verbose` flag will evaluate the env vars:
```
$ mule -f package-sign.yaml --list-env package-sign --verbose
deb:
- AWS_ACCESS_KEY_ID=foo
- AWS_SECRET_ACCESS_KEY=bar
- BRANCH=rel/stable
- CHANNEL=stable
- S3_SOURCE=$S3_SOURCE
- VERSION=2.2.0
```
In verbose mode, note that any env vars not defined in the calling environment will simply print the expected var name, as in the `S3_SOURCE` entry in the above example.

### DIFF
--- a/mule/driver.py
+++ b/mule/driver.py
@@ -16,51 +16,17 @@ def main():
         if len(mule_yamls) == 0:
             mule_yamls.append('mule.yaml')
         mule_config = _read_mule_yamls(mule_yamls)
-        parsed_mule_config = yaml_util.read_yaml(mule_config, raw=yaml_read_raw(args))
-        agent_configs, jobs_config, task_configs = validator.get_validated_mule_yaml(parsed_mule_config)
+        agent_configs, jobs_config, task_configs = _get_list_configs(mule_config, raw=yaml_read_raw(args))
+
         if args.list_agents:
-#<<<<<<< HEAD
-#            _, _, agent_configs = _get_list_configs(mule_config)
-#            _list_agents(agent_configs, args.verbose)
-#        elif args.list_env:
-#            job_name = args.list_env
-#            _list_env(mule_config, job_name, args.verbose)
-#||||||| merged common ancestors
-#            parsed_mule_config = yaml_util.readYaml(mule_config)
-#            _list_agents(parsed_mule_config["agents"], args.verbose)
-#=======
-#            _list_agents(agent_configs, args.verbose)
-#        elif args.list_jobs:
-#            _list_jobs(jobs_config, args.verbose)
-#        elif args.list_tasks:
-#            _list_tasks(task_configs, args.verbose)
-#>>>>>>> develop
             _list_agents(agent_configs, args.verbose)
         elif args.list_jobs:
             _list_jobs(jobs_config, args.verbose)
+        elif args.list_env:
+            _list_env(agent_configs, jobs_config, task_configs, args.list_env, args.verbose)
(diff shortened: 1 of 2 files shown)


mule/driver.py
```diff
@@ -34,6 +37,10 @@ def main():
         )
         raise error
 
+def _get_list_configs(mule_config, raw):
```
bricerisingalgorand: I like this helper function, but maybe calling this `_get_configs` since they're used both for the list options and the job executions. Or does the list here have different implications?

btoll: I'll change, there's no implication.

mule/driver.py
```diff
@@ -51,6 +58,49 @@ def _list_agents(agent_configs, verbose):
         else:
             print(agent['name'])
 
+def _list_env(agent_configs, jobs_config, task_configs, job_name, verbose):
```
bricerisingalgorand, line 62: I think we should maybe update our dict utils to have a function that searches for regex patterns in all entries in a map, then have this function just use that to print out items matching env var syntax. That way we can try to reuse the code we already have for traversing dictionaries

btoll: That may be useful in some scenarios.  For this, I think that it's starting to move away from the intent of this function.  I only want to get the `env` list in the agent block, and I don't want any other env vars defined in other keys, such as those that could be in `workdir` or `volumes`, for example.

I also like that this relies only on whatever has been explicitly defined in the `env` list, rather than something that could slip through a regular expression.

I believe that getting the `env` list is a good underlying feature to support tooling such as `saddle` that is for a specific purpose.

mule/driver.py
```diff
@@ -15,12 +17,14 @@ def main():
         if len(mule_yamls) == 0:
             mule_yamls.append('mule.yaml')
         mule_config = _read_mule_yamls(mule_yamls)
-        parsed_mule_config = yaml_util.read_yaml(mule_config, raw=yaml_read_raw(args))
-        agent_configs, jobs_config, task_configs = validator.get_validated_mule_yaml(parsed_mule_config)
+        agent_configs, jobs_config, task_configs = _get_configs(mule_config, raw=yaml_read_raw(args))
```
egieseke, line 20: Why not package as a single config object rather than returning 3 separate configs?

btoll: It's six of one, half dozen of the other, isn't it?

btoll: The functions below that line are expecting their respective configs, so that's why it has been returning the values this way.  If we were to return it as a single object, then each function would have to extract that info from the object.  Maybe not a big deal.

So do you return the separate configs from the callee or refactor to just return one object and then pass that into each function to be unpacked?

That's what I meant by the above comment.  Currently, I prefer the former as I think it reads better.

egieseke: OK

egieseke (approved):

onetechnical (approved):

PR #50, opened 2020-11-09
Improve error messaging to include possible missing env vars for clarity

btoll:
Currently, the error message for BucketCopy tasks should be clarified when not including needed variables.  For instance, it's not clear that that is the issue.  Here is the current error message:

```
Mule could not execute job stage in file ['mule.yaml']
Source and destination configs cannot both be local
```

Here is the improved message:

```
Mule could not execute job stage in file ['mule.yaml']
Source '' and destination '' configs cannot both be local
Are all needed environment variables set?
```

### DIFF
--- a/mule/task/s3/__init__.py
+++ b/mule/task/s3/__init__.py
@@ -2,6 +2,7 @@ import re
 import os
 import pathlib
 
+from mule.error import messages
 from mule.task import ITask
 from mule.util import s3_util
 
@@ -173,7 +174,7 @@ class BucketCopy(ITask):
                 # To get the wildcard, join the last two tuple elements, i.e., ('', 'bar, '*.out') => 'bar/*.out'
                 s3_util.upload_files('/'.join((src_bucket, src_prefix)), dest_bucket, dest_prefix)
         else:
-            raise ValueError('src and dest configs cannot both be local')
+            raise ValueError("\n".join((messages.BUCKET_UPLOAD_SRC_DEST, messages.POSSIBLE_MISSING_ENV_VARS)).format(src_bucket, dest_bucket))
 
     def execute(self, job_context):
         super().execute(job_context)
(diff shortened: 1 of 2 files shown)


mule/task/s3/__init__.py
```diff
@@ -173,7 +174,7 @@ def copy(self):
                 # To get the wildcard, join the last two tuple elements, i.e., ('', 'bar, '*.out') => 'bar/*.out'
                 s3_util.upload_files('/'.join((src_bucket, src_prefix)), dest_bucket, dest_prefix)
         else:
-            raise ValueError('src and dest configs cannot both be local')
+            raise ValueError("\n".join((messages.BUCKET_UPLOAD_SRC_DEST, messages.POSSIBLE_MISSING_ENV_VARS)).format(src_bucket, dest_bucket))
```
egieseke, line 177: That seems odd to say that env vars may be missing.  How can that occur?

btoll: It's very easy to do:

```
mule -f mule.yaml stage
```

We can't enforce env vars are present for some tasks.  All we can do in these cases is try to give the user a hint.

egieseke: Can we print out the environment variables are if we suspect that they are wrong?

btoll: It's already doing that.

btoll: What's wrong with giving the user a hint?  The current message isn't helpful when the user enacts the task without the proper env vars.

egieseke: I am all for messages that help clarify the error.  The wording seems a little odd to me and less than clear.

egieseke (approved):

mule/task/s3/__init__.py
onetechnical: It seems weird to use two different error message handlers in the same error message, but it's probably fine.

onetechnical (approved):

mule/task/s3/__init__.py
btoll: The intent is to have `POSSIBLE_MISSING_ENV_VARS` as a generic error message that can be appended or used singly wherever it's needed.

PR #51, opened 2020-11-20
Added `destroy` config to make auto-destroying a container optional

btoll:
This will be helpful when using agents that will be used to create
snapshots of our package repos.

After all, switching out existing snapshots created from a previous run
of the container is much faster than publishing a snapshot from a new
local cache.

### DIFF
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -18,6 +18,7 @@ class Docker(ITask):
     machine = {
         'arch': 'amd64',
         'buildArgs': [],
+        'destroy': True,
         'dockerFilePath': 'Dockerfile',
         'env': [],
         'image': '',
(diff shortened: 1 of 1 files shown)


mule/task/docker/__init__.py
```diff
@@ -118,17 +113,19 @@ def pullFromDockerHub(self, image):
             found = True
         return found
 
-    def run(self, image, command, work_dir, volumes, env):
+    def run(self, image):
```
btoll, line 116: This function signature was beginning to get unwieldy.

onetechnical (approved):
I haven't run it but these changes look fine to me.

egieseke (approved):
Looks ok.  

Food for thought, I wonder if we can simplify our docker interaction using the docker SDK. 
https://docker-py.readthedocs.io/en/stable/

btoll:
> Looks ok.
> 
> Food for thought, I wonder if we can simplify our docker interaction using the docker SDK.
> https://docker-py.readthedocs.io/en/stable/

Probably be worth looking into.

PR #53, opened 2020-11-30
Added command substitution to agent `env` values

btoll:
## Summary

Added command substitution to agent `env` values.  This brings feature parity with the `buildArgs` config.

- fixed bug in subprocess.run (create an array first)
- changed function name format to be pythonic

This continues to get us closer to full automation as allowing for command substitution in "env" will potentially allow for the removal of some environment variables that are computed at runtime in the task shell scripts.

## Test Plan

Apply this patch:
```
diff --git a/Makefile b/Makefile
index 0966d019..8132f97d 100644
--- a/Makefile
+++ b/Makefile
@@ -308,3 +308,6 @@ archive:
        CHANNEL=$(CHANNEL) \
        aws s3 cp tmp/node_pkgs s3://algorand-internal/channel/${CHANNEL}/$(FULLBUILDNUMBER) --recursive --exclude "*" --include "*${CHANNEL}*$(FULLBUILDNUMBER)*"
 
+hello:
+       echo hello world!
+
diff --git a/package.yaml b/package.yaml
index d5d8de0d..e5adf0ec 100644
--- a/package.yaml
+++ b/package.yaml
@@ -17,8 +17,10 @@ agents:
     buildArgs:
       - GOLANG_VERSION=`./scripts/get_golang_version.sh`
     env:
-      - NETWORK=$NETWORK
-      - VERSION=$VERSION
+      - NETWORK=mainnet
+      - VERSION=`echo 2.1.1`
+      - ARCH_TYPE=`uname -m`
+      - OS_TYPE=`./scripts/ostype.sh`
     workDir: $HOME/projects/go-algorand
 
   - name: docker
@@ -55,7 +57,16 @@ tasks:
     agent: docker
     target: mule-package-docker
 
+  - task: docker.Make
+    name: hello
+    agent: rpm
+    target: hello
+
 jobs:
+  hello:
+    tasks:
+      - docker.Make.hello
+
   package:
     tasks:
       - docker.Make.build
```
Then, run `mule -f package.yaml hello`.

### DIFF
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -88,6 +88,11 @@ class Docker(ITask):
 
     def execute(self, image):
         self.ensure(image)
+        validations = [
+                ('env', '=', re.compile(r'.*=.+')),
+                ('volumes', ':', re.compile(r'.+:.+'))]
+        for field, delimiter, regex in validations:
+            self.validate_configs(field, delimiter, regex)
         self.run(image)
 
     def eval_args(self, args_list):
@@ -123,9 +128,9 @@ class Docker(ITask):
         container_name = f"mule-{time.time_ns()}"
         docker_command = ['docker', 'run', '--name', container_name, rm_container, '-w', work_dir, '-i', '-v', f"{os.getcwd()}:{work_dir}"]
 
-        for env_var in self.eval_args(self.machine['env']):
+        for env_var in self.machine['env']:
             docker_command.extend(['--env', env_var])
-        for volume in self.validate_volumes(self.machine['volumes']):
+        for volume in self.machine['volumes']:
             docker_command.extend(['-v', volume])
 
         docker_command.append(image)
@@ -133,12 +138,31 @@ class Docker(ITask):
         atexit.register(self.kill, container_name)
         subprocess.run(docker_command, check=True)
 
-    def validate_volumes(self, volumes):
-        volume_re = re.compile(r'.+:.+')
-        bad_volumes = [volume for volume in volumes if not volume_re.match(volume)]
-        if len(bad_volumes):
-            raise Exception(messages.BAD_VOLUME_CONFIG.format(", ".join(bad_volumes)))
-        return volumes
+    def validate(self, field, re):
+        try:
+            if not re.match(field):
+                raise ValueError
+            return True
+        except TypeError:
+            raise Exception(messages.AGENT_FIELD_WRONG_TYPE.format(
+                self.machine['name'],
+                field,
+                str,
+                type(field)
+            ))
+        except ValueError:
+            raise Exception(messages.AGENT_FIELD_WRONG_FORMAT.format(
+                self.machine['name'],
+                field
+            ))
+
+    def validate_configs(self, field, delimiter, regex):
+        if type(self.machine[field]) is dict:
+            self.machine[field] = [f"{k}{delimiter}{v}" for k, v in self.machine[field].items()]
+        elif type(self.machine[field]) is list:
+            self.machine[field] = [f for f in self.machine[field] if self.validate(f, regex)]
+        else:
+            raise Exception("Unsupported type")
 
 
 class Make(Docker):
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -125,13 +125,20 @@ class Docker(ITask):
 
         for env_var in self.eval_args(self.machine['env']):
             docker_command.extend(['--env', env_var])
-        for volume in self.machine['volumes']:
+        for volume in self.validate_volumes(self.machine['volumes']):
             docker_command.extend(['-v', volume])
 
         docker_command.append(image)
         docker_command.extend([self.machine['shell'], '-c', self.command])
         atexit.register(self.kill, container_name)
         subprocess.run(docker_command, check=True)
+
+    def validate_volumes(self, volumes):
+        volume_re = re.compile(r'.+:.+')
+        bad_volumes = [volume for volume in volumes if not volume_re.match(volume)]
+        if len(bad_volumes):
+            raise Exception(messages.BAD_VOLUME_CONFIG.format(", ".join(bad_volumes)))
+        return volumes
 
 
 class Make(Docker):
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -10,6 +10,8 @@ from mule.task import ITask
 from mule.error import messages
 from mule.util import update_dict
 
+# TODO: isolate all `subprocess.run` calls to a single location
+
 class Docker(ITask):
     required_fields = [
         'image',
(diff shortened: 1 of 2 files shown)


mule/task/docker/__init__.py
```diff
@@ -129,28 +132,10 @@ def run(self, image):
         atexit.register(self.kill, container_name)
         subprocess.run(docker_command, check=True)
 
-    def validateDockerConfigs(self):
-        for env_var_index, env_var in enumerate(self.machine['env']):
-            if not type(env_var) == str:
```
btoll, line 134: Aren't these always going to be strings?  ( Same as below for the volumes. )

mule/task/docker/__init__.py
```diff
@@ -129,29 +131,6 @@ def run(self, image):
...
-                raise Exception(messages.TASK_FIELD_IS_WRONG_TYPE.format(
-                    self.getId(),
-                    f"docker.env[{env_var_index}]",
-                    str,
-                    type(env_var)
-                ))
-        dockerEnvVarPattern = re.compile(r'.*=.+')
-        self.machine['env'] = [envVar for envVar in self.machine['env'] if dockerEnvVarPattern.match(envVar)]
```
btoll, line 142: Checking "env" and "volumes" below is potentially masking bad configs, since ones not matching the regex will be discarded.

I think it'd be better to let it explicitly fail so the user has an idea of how to fix it.

onetechnical (commented):
I think overall this is fine, but want others to review/approve this.

mule/task/docker/__init__.py
bricerisingalgorand: They should be, but yaml will allow someone to incorrectly put an int or boolean into the env list. This is more of a failsafe than anything in preparation for checking that they match the regex pattern

mule/task/docker/__init__.py
bricerisingalgorand: That makes sense, throwing an exception would be a good idea

mule/task/docker/__init__.py
```diff
@@ -86,26 +88,28 @@ def ensure(self, image):
 
     def execute(self, image):
         self.ensure(image)
-        self.validateDockerConfigs()
```
bricerisingalgorand, line 89: We need to keep this part to validate that the yaml configs are syntactically configs, unless this is now implemented somewhere else. Otherwise things the docker run command will fail without an explanation

mule/task/docker/__init__.py
btoll: We don't want to check for only strings (although I'm not sure I understand how anything other than a string could be passed) since this ticket is allowing for command substitution to evaluate the value of an environment variable.

mule/task/docker/__init__.py
btoll: We should let docker fail, though, instead of us trying to anticipate by throwing our own exception.

mule/task/docker/__init__.py
btoll: I don't think that we should be doing these checks, though, for a couple of reasons.  First, the validation for the env var is obsoleted now that we're also allowing for command substitution.  Second, if any bad params are passed to the docker run or build command, we should let docker throw an error.  I don't think that we should try to anticipate any errors, and docker is already well-vetted to do so.

Also, the validation that we're doing is more masking, which can be confusing for the user.  It's better for the command to fail outright, and docker will throw its own error message that we are printing.

mule/task/docker/__init__.py
btoll: I've been thinking on this, and I agree with you that I think it's worth throwing an exception if there isn't a colon in the string pattern.  That would be a more helpful error message.

onetechnical (dismissed):

egieseke (commented):
I will approve once the conversation with Brice is concluded.

mule/task/docker/__init__.py
bricerisingalgorand: But the docker module won't understand why there was a problem. These checks are done to try and guarantee that the yaml schema is met, if we don't do that then we're increasing the complexity of the error handling in all of our downstream modules since the scope of problems won't be bounded by just incorrect parameters. I agree with improving the schema validations and giving more user feedback, but removing schema validations as a whole is just dangerous

mule/task/docker/__init__.py
bricerisingalgorand: The goal of this is to guarantee that the yaml written for this component matches our schema. By doing this the errors that we can potentially have are at least bounded by the parameters we make available to users through our yaml language. If we don't do that here, then downstream modules will have a lot more errors they need to handle and won't be able to inform the user that their parameters were incorrect since they won't know anything about the yaml schema

btoll: I added this yesterday after leaving that comment: https://github.com/algorand/mule/pull/53/commits/24d78ab07514acab4ef88882cd865aa394935b7f

mule/task/docker/__init__.py
btoll: The checks that I've removed are that the env vars and the volumes are strings.  Now, that we're allowing for command substitution, I believe that that obsoletes the string check for the env var.  But even if we weren't allowing for command substitution, the type(str) checks are redundant.  Evaluating the env vars will give us strings.

I don't think we really have a schema as much as we're massaging inputs into forms that `docker build` and `docker run` will accept.  Let docker fail if given bad inputs, just as it would on the cli.

mule/task/docker/__init__.py
```diff
@@ -129,28 +138,31 @@ def run(self, image):
...
-            if not type(volume) == str:
-                raise Exception(messages.TASK_FIELD_IS_WRONG_TYPE.format(
-                    self.getId(),
-                    f"docker.env[{volume_index}]",
-                    str,
-                    type(volume)
-                ))
-        dockerVolumePattern = re.compile(r'.+:.+')
-        self.machine['volumes'] = [volume for volume in self.machine['volumes'] if dockerVolumePattern.match(volume)]
+    def validate(self, field, re):
```
bricerisingalgorand, line 141: Hmm maybe this could be a util function since it's just matching regex and throwing certain errors. That might be useful for other tasks in the future. Maybe something like this:

```python
def validateField(dict, key, re):
    try:
       matches = re.match(dict[key])
       if not matches:
            raise ValueError
       return matches
    except KeyError:
         raise Exception("some key not found error")
    except ValueError:
         raise Exception(messages.FIELD_VALUE_INVALID.format(
                key,
                dict[key],
                re
            )
```

I think it works the way it is though, that might just be an idea for the future

bricerisingalgorand (approved):

mule/task/docker/__init__.py
btoll: Yes, I think this works for now.

I think it's better,though, to have all schema validation happen in the library and not the endpoints.  If I'm writing a module/plugin for mule, I feel it's reasonable to expect that what I'll be handed is a context object containing data that's already been validated.  Then, I just have to worry about assembling the pieces into ways that make sense to my particular module, in this case massaging the data to fit `docker build` and `docker run`.

It's one of the many benefits of having a library.

egieseke (approved):
Looks good.

PR #59, opened 2020-12-09
--list-env: Check for agent in task config

btoll:
## Summary

The implementation of the `--list-env` command wasn't properly checking for any defined agents in a task, and this led to an unseemly error.

We now check for that and print a helpful message.

## Test Plan

Run any `mule` job that references a task without a defined agent.


onetechnical (approved):

egieseke (approved):
Looks good.

PR #60, opened 2020-12-12
Add recipe support

btoll:
## Summary

`mule` now supports passing in recipe files on the command line.  An example can look like this:
```
mule --recipe sign_and_test.yaml
```
or, from `stdin`:
```
curl https://remote-resource.recipe | mule -r -
```

Note that it is not necessary to pass a reference to a mule yaml file.  This is because the recipe, generated by the `saddle` plugin, includes all of the state information needed to run the job(s).

`mule` will forward the file stream to the `saddle` plugin, which will take the recipe file and return a "compiled" file that `mule` can then use to run its tasks.  Since the compiled file includes all of the state for all of the jobs and tasks, it is not necessary to further reference the mule yaml file(s) from which the recipe was built.  This will help simplify any downstream tooling.

For reference, a recipe file will take the following form:
```
created: 2020-12-17 22:45:44.602634
filename: /home/btoll/projects/go-algorand/package_sign_and_test.yaml
mule_version: 0.0.13
env:
  BRANCH: rel/stable
  NETWORK: mainnet
  S3_SOURCE: algorand-staging/releases
  SHA: 07bc97b7b58b8f197cd96919639765ab1855cab1
  VERSION: 2.3.0
jobs:
- package-sign
- package-test-deb
- package-test-rpm
```

As a result of that, I needed to update `mule` to accommodate that.  Note that `mule` only supports running multiple sequential jobs for recipes, not for the more common use case (and prior to this PR, the **only** use case).

In addition, the following changes were made:

1. Added `_get_job_config` to `driver.py`.  This function will get all the bits needed for a particular job, and its returned dictionary will be used as the job's context (`JobContext`).  This is important and will allow for any subsequent operations to only operate on the job's context and not on the whole `mule` config (or a part of it, such as `tasks`).  Constraining the state to be operated on or inspected will help to limit the scope of the work state and to help limit bugs that can occur by reaching into objects either shared or owned by others.

1. Lifted the validation of the `agent`'s `env` and `volumes` blocks out of the `docker` module and into the library's `validator` module.  This was necessary allow for the `saddle` plugin to operate on `dict`s instead of `list`s, which is much easier to do (for example, it's easier to work on `{ "version": "2.3.0" }` than `[ "version=2.3.0" ]`.  Schema validation is best left up to the library anyway, and removing it from the `docker` module simplifies it.

Many of the changes on this PR are the results of running the touched files through `pycodestyle`.

## Test Plan

This has been tested extensively and supports all current `mule` functionality.  There are no breaking changes.


egieseke (commented):
Once an update to use a simplified recipe is incorporated, I will approve.

mule/task/__init__.py
```diff
@@ -89,39 +86,36 @@ class Job(ITask):
         ('configs', dict)
     ]
 
-    def __init__(self, args):
-        super().__init__(args)
-        self.dependencies = args['tasks']
-        self.task_configs = args['task_configs']
-        self.job_configs = args['configs'] if 'configs' in args else {}
-        if 'agent_configs' in args:
-            self.agent_configs = args['agent_configs']
```
btoll, line 98: The `task_configs` and `job_configs` can be looked up in job context.

mule/util/__init__.py
```diff
@@ -1,20 +1,18 @@
...
 
-class JobContext:
-
-    job_context = {}
 
-    def __init__(self, agent_configs):
-        self.job_context['agents'] = agent_configs
+class JobContext:
+    def __init__(self, job_config):
+        self.job_context = copy.deepcopy(job_config)
```
btoll, line 8: `job_context` should now be an instance property.

setup.py
```diff
@@ -32,6 +32,12 @@
...
         "console_scripts": [
             "mule = mule.driver:main"
-        ]
+        ],
+        "mule.plugin": [
+            "docker = mule.task.docker",
+            "s3 = mule.task.s3",
+            "shell = mule.task.shell",
+            "stash = mule.task.stash",
+        ],
```
btoll, line 41: Setting this up here and exposing as plugins in anticipation of looking up the referenced modules by group name rather than by `pydoc.locate`.

mule/driver.py
```diff
@@ -15,43 +15,91 @@ def main():
...
+            out = saddle.get_compiled(args.recipe)
+            job_yaml = yaml.safe_load(out)
+            job_configs = yaml_util.read_yaml(job_yaml.get("items"), raw=False)
+            for j_c in job_configs:
+                _execute_job(j_c)
+            return
+
+        mule_config = _get_configs(
+                _read_mule_yamls(mule_yamls),
+                raw=yaml_read_raw(args))
```
onetechnical, line 36: Oh neat, we combined these. 👍

mule/driver.py
```diff
@@ -15,43 +15,95 @@ def main():
...
     try:
         mule_yamls = args.file
-        if len(mule_yamls) == 0:
+        if not len(mule_yamls):
             mule_yamls.append('mule.yaml')
-        mule_config = _read_mule_yamls(mule_yamls)
-        agent_configs, jobs_config, task_configs = _get_configs(mule_config, raw=yaml_read_raw(args))
+
+        if args.recipe:
+            plugins = mule.validator.get_plugin("saddle")
```
bricerisingalgorand, line 22: I think this is ok for now, but if we're going to have plugin support maybe we should support them more generically. We could support plugin names as alternative arguments to jobs. In this case we could support it like so: `mule saddle -r recipe.yaml`

Maybe we could have all plugins take parameters from the command and provide the compiled job_context to mule that gets run.

mule/error/messages.py
```diff
@@ -17,6 +17,7 @@
 MACHINE_MISSING_NAME="Could not find machine name '{}'"
 MULE_DRIVER_EXCEPTION="Mule could not execute job {} in file {}\n{}\n"
 POSSIBLE_MISSING_ENV_VARS="Are all needed environment variables set?"
+PLUGIN_NOT_FOUND="The `{}` plugin was not found"
```
bricerisingalgorand, line 20: Maybe some context around the activity that required the plugin would be good here

mule/task/__init__.py
```diff
@@ -4,49 +4,47 @@
...
-class ITask:
 
+class ITask:
     required_fields = []
     required_typed_fields = []
     optional_typed_fields = []
     dependencies = []
 
-    def __init__(self, args):
+    def __init__(self, job_config):
```
onetechnical: It feels weird to instantiate a task with a job config, but I guess it is consolidated by this point. Should probably be documented.

mule/validator/__init__.py
```diff
@@ -154,12 +224,12 @@ def getValidatedTaskDependencyChain(job_context, dependency_edges):
     tasks_tbd.reverse()
     return tasks_tbd
 
-def validateRequiredTaskFieldsPresent(task_id, fields, required_fields):
+
+def validate_required_tasks_fields_present(task_id, fields, required_fields):
```
onetechnical: Meant to change the function name from "required task fields" to "required tasks fields"?

onetechnical (dismissed):
I had some comments, but my quick look says this is okay.

mule/driver.py
bricerisingalgorand, line 22: Maybe we should have this method only return one plugin and throw an error if it finds multiple, since mule can't really know which plugin to use of there are duplicates right?

mule/driver.py
```diff
@@ -15,43 +15,95 @@ def main():
...
+    return {
+        "name": job,
+        "configs": job_configs,
+        "agents": agents,
+        "tasks": tasks,
+        "task_configs": task_configs,
+    }
+
+
+def _get_task(mule_config, job_task):
```
bricerisingalgorand, line 98: I think this is unnecessary since when you instantiate tasks they figure out their own id. Unless there's a reason we need to change the structure of the task_configs object

mule/driver.py
```diff
@@ -15,43 +15,95 @@ def main():
...
-    return mule_config
+
+def _get_job_config(mule_config, job):
+    job_def = mule_config["jobs"].get(job)
+    job_configs = job_def.get("configs", {})
+    tasks = job_def.get("tasks", [])
+    task_configs = []
+    agents = []
+    for job_task in tasks:
+        name, task = _get_task(mule_config, job_task)
```
bricerisingalgorand, line 79: Why not include all task configs, since the tasks to run will be listed?

mule/driver.py
```diff
@@ -60,18 +112,19 @@ def _list_agents(agent_configs, verbose):
...
-        tasks = jobs_config[job_name]["tasks"]
+        tasks = mule_config["jobs"][job_name]["tasks"]
         # Different tasks can have the same agent, so let's use a set.
         agents = set()
 
         for task in tasks:
             # Now that we have each task, loop over the yaml task configs (i.e., `tasks`)
             # to determine the agent defined for the specified tasks.
-            for t_c in task_configs:
+            for t_c in mule_config["tasks"]:
```
bricerisingalgorand: We should use better variable names for these, task_config and agent_config instead of t_c and a_c will be a lot easier reading

mule/task/__init__.py
bricerisingalgorand: Requiring tasks to read the job_config object is a little dangerous since changes there will require changes here as well. We should upstream provide just the task config to this class

mule/driver.py
```diff
@@ -15,43 +15,95 @@ def main():
...
-    mule_config = {}
-    for mule_yaml in mule_yamls:
-        update_dict(
-            mule_config,
-            file_util.readYamlFile(mule_yaml),
-            overwrite_lists=False
-        )
-    return mule_config
+
+def _get_job_config(mule_config, job):
```
bricerisingalgorand, line 72: I think it would be better for us to create a Job object here and implement a get_config() method in Job. That way the Job class is responsible for declaring and providing configurations it can read/knows how to use. This way the code will be easier to maintain since the driver module won't need to know anything about the structure of job configs

mule/task/docker/__init__.py
```diff
@@ -46,6 +46,14 @@ def build(self, image):
         docker_command = f"docker build {build_args_str} {tags_str} -f {self.machine['dockerFilePath']} {self.machine['context']}"
         subprocess.run(docker_command.split(' '), check=True)
 
+    # The mule schema supports defining agent.{env,volumes} as either dicts or lists,
+    # but we want to operate on lists in this module because it's easier to build out
+    # the docker api.
+    def check_block_format(self, l, delimiter):
```
bricerisingalgorand, line 52: The naming of this function is a little confusing. Something like "convert_dict_to_list" might be better

mule/task/docker/__init__.py
```diff
@@ -128,42 +131,17 @@ def run(self, image):
...
+
+        for volume in self.check_block_format(self.machine['volumes'], ":"):
             docker_command.extend(['-v', volume])
 
         docker_command.append(image)
         docker_command.extend([self.machine['shell'], '-c', self.command])
         atexit.register(self.kill, container_name)
         subprocess.run(docker_command, check=True)
 
-    def validate(self, field, re):
```
bricerisingalgorand, line 141: I thought we talked about why we need this the other week. We still need to guarantee that the parameters passed are all the correct type/structure

bricerisingalgorand: Oh i see you moved it elsewhere

mule/validator/__init__.py
```diff
@@ -14,27 +16,84 @@
...
+validators = {
+    "env": re.compile(r".*=.+"),
+    "volumes": re.compile(r".+:.+")
+}
+
+
+def get_plugin(name):
+    return [
+        entry_point.load()
+        for entry_point in pkg_resources.iter_entry_points(group="mule.plugin")
```
bricerisingalgorand, line 29: If we're going to manage plugins this way, we should add it as another config option in the mule config

mule/validator/__init__.py
```diff
@@ -14,27 +16,84 @@
 
 DEFAULT_MULE_CONFIG_PATH = "~/.mule/config.yaml"
 
+
+validators = {
```
bricerisingalgorand, line 20: I think putting a full yaml schema would be appropriate here, but not just validation regex for agent objects. Instantiating these in validate_agent_block would probably be ok for now

mule/validator/__init__.py
```diff
@@ -14,27 +16,84 @@
...
     if os.path.isfile(config_file_path):
-        mule_configs_from_file = file_util.readYamlFile(config_file_path)
+        mule_configs_from_file = file_util.read_yaml_file(config_file_path)
         mule_configs.update(mule_configs_from_file)
     return mule_configs
 
-def get_validated_mule_yaml(mule_config):
-    # Note that agents are optional!
 
+def validate(field, f):
```
bricerisingalgorand, line 43: I think what this method really does is field_matches_regex right? I don't think trying to generically validate fields based on regex is going to work since different fields take different types. I would have this module just validate types, then let classes using the configs validate things like regex, number bounds, etc

mule/driver.py
btoll: Yeah, I agree this should be talked through.  I left this as-is for now because I didn't want to create anything now that may not work in the future (like a separate module, different ways of stating the use of the plugin to use on the cli, etc).

mule/task/__init__.py
btoll: This change was just to make the code more understandable, i.e. moving from the ambiguous `args` to the explicit `job_config`.

I had thoughts of having a job context singleton that wouldn't need to be passed, but it would create too much churn.

mule/validator/__init__.py
btoll: Oh snap, great catch.  No, I didn't :).  I'll double-check all that and rename it back to what it was.

mule/driver.py
btoll: Yes, there is no support for multiple plugins currently.

I think this is fine for now to get this to work.  We'll all need to decide on what the plugin interface should look like.

mule/driver.py
btoll: Will change.

mule/task/__init__.py
btoll: The only change here is to the variable name for better readability.

mule/driver.py
btoll: I have a lot of thoughts on this area too on how to improve it and have it be better accessible to plugins that need to get their job config(s), but I don't want to create a lot of architecture around that before we all talk about how the plugin architecture should be designed.

mule/validator/__init__.py
btoll: Yes, I agree, this is a good first step.  Designing this properly would have greatly increased the scope of this ticket.

mule/validator/__init__.py
btoll: This is a good first step.  Designing this properly will require more thought and the changes will/would be too great for this PR.

btoll:
@bricerisingalgorand Keep in mind that this ticket is only scoped to bring recipe support to `mule`.  Your suggestions are good, but I think they are out of scope for this ticket, as doing them will be a lot of refactoring.

It'd be a good idea to have a sysarch meeting just devoted to `mule` where we can all start fleshing out the areas that need work, such as schema validation, plugin architecture, job contexts, etc.

My changes here to support the `saddle` plugin are meant to be easy to change, as this area of `mule` will be bound to change (plugin support).  I feel that anything more than what I did here would be going down the road of changes to the whole package which would be harder to further change and reason about once the dust settles.

mule/task/docker/__init__.py
btoll: Did you read the PR summary? :)

I described my reasons to do so there.

mule/driver.py
btoll: I wanted the job context to only include the tasks that they need to know about to run.  Otherwise, global state creeps in and leads to bugs.

Since we're supporting multiple sequential jobs via the `saddle` plugin (and this support is only bound to grow for other use cases as well), each one should have its own copy of the context.

One area where this was a problem was copying `configs` into all the tasks where the `configs` were overriding those defined in the task.  When each job doesn't have its own copy of the world, this leads to hard to find bugs because they all have the same reference to the task(s).

mule/driver.py
btoll: See my comment below.

mule/driver.py
btoll: Yeah, creating a job context necessitated this.

mule/driver.py
btoll: Also, I'm not changing the structure.  There aren't any breaking changes on this ticket.

mule/task/docker/__init__.py
btoll: Well, it's not always going to convert.  Also, if we support other types such as tuples we'd have to change the name.  I think it's better to leave this defined by what it does, which is checking blocks of data rather than specific types.

mule/error/messages.py
btoll: Good idea, how about we decide when we talk further about plugin architecture?

mule/validator/__init__.py
btoll: Please see my other comments on this matter.

mule/task/__init__.py
onetechnical: I kind of echo what @bricerisingalgorand says here and would prefer to pass the task config semantically, but don't feel strongly on it.

onetechnical (dismissed):

bricerisingalgorand (approved):

onetechnical (approved):

PR #65, opened 2021-01-01
Added bash completion to mule cli

btoll:
## Summary

The goal of this is to add bash tab completion to the `mule` cli.  This is a nice feature and is expected of any production-ready cli tool.

There were a number of changes that needed to be made in order for this to work properly.
1.  Only print function results from within the `main` function.  Helper functions such as `list_{agents,jobs,tasks}` should not print but instead return the results to the caller.  The aforementioned functions are now are part of the public API so they can be used by any application that imports `mule`.
1.  Renamed `driver.py` -> `main.py` and moved code out of `main.py` into new `mule.py` module.
1.  This will then allow for the following:
    - import `mule` as a module
    - tooling to work more seamlessly (callers can work with results instead of having to clumsily read from `stdin`)
    - much easier to unit test
1.  Implement a logger.  This allows for granular control over any logger(s) that we create, including formatting, levels, handlers, etc.  We need better control over what is logged and where.
1.  Removed the `termcolor` dependency in favor of `colorlog`.  The latter is implemented to work with the python logging system.
1.  Removed any "args.*" pass-through variables.  This will keep all invocations from the cli in the `main()` function.

There are no breaking changes.

## Test Plan

Tested in `saddle` package.  Removed all `mule` subprocess commands and instead imported `mulecli` in as a dependency.
Manually tested.


mule/util/yaml/env_var_loader.py
```diff
@@ -9,15 +9,7 @@
...
-    value = os.path.expandvars(node.value)
-    if env_var_path_matcher.match(value):
-        cprint(
-            messages.CANNOT_EVALUATE_ENV_VAR.format(node.value),
-            'yellow',
-            attrs=['bold']
-        )
-        return ''
-    return value
+    return os.path.expandvars(node.value)
```
btoll, line 12: This code has been removed because it is making the tooling difficult to do.  Also, there is a very good use case why the env vars wouldn't be defined in the host environment.

This isn't a bad check, but I think it should be able to be turned on or off, and I think the default should be it's off.  We could control the level of verbosity in a number of different ways.

mule/driver.py
```diff
@@ -3,12 +3,13 @@
...
-import mule.validator as validator
-import mule.util.yaml.env_var_loader as yaml_util
-from mule.util import JobContext, file_util, prettify_json, update_dict
 from mule.error import messages
+from mule.logger import logger
 from mule.task import Job
+from mule.util import JobContext, file_util, prettify_json, update_dict
+import mule.parser
+import mule.util.yaml.env_var_loader as yaml_util
+import mule.validator as validator
```
btoll: This noise is mostly just sorting.  The only thing new is importing the logger.

mule/parser.py
```diff
@@ -59,7 +64,6 @@ def parseArgs():
     )
 
     group.add_argument(
-        '-l',
```
onetechnical, line 62: Why did we remove this?

mule/task/__init__.py
```diff
@@ -13,7 +13,6 @@ class ITask:
     dependencies = []
 
     def __init__(self, task_config):
-        ipdb.set_trace()
```
onetechnical, line 16: Whoops, leftover debug statement? Nice catch.

mule/validator/__init__.py
```diff
@@ -18,7 +18,7 @@
 
 
 validators = {
-    "env": re.compile(r".*=.+"),
+    "env": re.compile(r"[a-zA-Z_][a-zA-Z0-9_]+=[a-zA-Z_$][a-zA-Z0-9_]+"),
```
onetechnical: Hmm, much tighter restrictions here...no longer allowing single character envs I see. All of this is fine for our purposes, just interesting restrictions.same on the other side, can't be single character, which is probably fine. (Changing the second character classes to * instead of + would fix that, but not sure if we care.) Maybe `[a-zA-Z_][a-zA-Z0-9_]*=$?[a-zA-Z_][a-zA-Z0-9_]*` for example.

mule/logger.py
```diff
@@ -0,0 +1,66 @@
...
+    "CRITICAL": "red",
+}
+
+log_date_format = "%Y-%m-%d %H:%M:%S"
+log_format = "[%(asctime)s] %(levelname)-8s %(name)-12s %(message)s"
+stream_format = "%(log_color)s%(levelname)-4s%(reset)s %(message)s"
+
+
+def setup_file_logger(logger, tempfile):
+    fileHandler = logging.FileHandler(tempfile)
```
onetechnical: Why not snake case?

mule/logger.py
```diff
@@ -0,0 +1,66 @@
...
+        datefmt=log_date_format,
+        reset=True,
+        log_colors=colors
+    )
+    logger.addHandler(fileHandler)
+
+
+def setup_logger():
+    logger = logging.getLogger(__name__)
+    setup_stream_logger(logger)
```
onetechnical, line 17: Stylistic question: why not just insert setup_stream_logger here?

mule/logger.py
```diff
@@ -0,0 +1,66 @@
...
+            log_colors=colors
+        )
+    )
+    logger.addHandler(console_handler)
+    logger.setLevel(logging.INFO)
+
+
+def start_debug(args):
+    utc_time = datetime.utcnow().isoformat()
+    tempfile = f"/tmp/mule.{utc_time}.log"
```
onetechnical: Hmmm...okay for now. I feel like we need better tempfile usage here, like using the tempfile module. That, or don't datestamp the logs so that we can later implement proper log rotation.

mule/util/__init__.py
```diff
@@ -38,7 +38,7 @@ def prettify_json(raw_json):
     return json.dumps(raw_json, indent=4, sort_keys=False)
 
 
-def update_dict(current, new, overwrite_lists=True):
+def parse_mule_file(current, new, overwrite_lists=True):
```
onetechnical: Interesting rename...I guess it's alright because it's only called in the mule_config instantiation, but update would imply that it's mutating the file, as opposed to parse which implies just reading.

onetechnical (commented):
I had some stylistic questions, but overall I like the approach here. It _seems_ okay to me, but want to defer to others to review.

mule/parser.py
btoll: Frankly, I think it was a mistake to add it.  There are several `--list-*` flags, and it's weird that only one would be referenced by the `-l` shorthand.

The bash completion (hopefully) makes this moot.

mule/logger.py
btoll: Ah, I'll change, it should be.

mule/logger.py
btoll: I wanted to keep the concerns separated.

mule/util/__init__.py
btoll: I'm open to a better name, but `update_dict` is way too generic.

I'd like to have a better implementation, since it is mutating the `current` object.  I'd rather it not mutate any state.

mule/logger.py
btoll: I didn't want to use the `tempfile` module because file creation (and closing, hopefully) is delegated to the `logging` module.

mule/validator/__init__.py
btoll: Ah, yeah, this was a mistake.  That was meant to only be for the name.  I'll change.

https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html

btoll: Fixed.

mule/logger.py
btoll: Fixed.

mule/util/__init__.py
onetechnical: Where it is I think it's fine to mutate the object, because it's creating it. `update_mule_config` might make more sense.

btoll: I'll change.

btoll: Fixed.

egieseke (dismissed):
Command completion will be helpful.  Log changes and refactoring of file names should probably be in separate PRs.

btoll:
> Command completion will be helpful. Log changes and refactoring of file names should probably be in separate PRs.

Separate PRs is a good idea, when we get better at timely reviews I'll start doing that more.

mule/logger.py
bricerisingalgorand: I'm confused is the implication here that logs when running in debug will go to a file?

mule/main.py
```diff
@@ -0,0 +1,112 @@
...
+                return
+
+        if args.debug:
+            start_debug(args)
+
+        mule_config = get_configs(mule_yamls, raw=_yaml_read_raw(args))
+
+        if args.list_agents:
+            if args.verbose:
+                print(prettify_json({agent.get("name"): agent for agent in mule_config.get("agents")}))
```
bricerisingalgorand, line 89: Should we be using logger here instead of print?

bricerisingalgorand: Also should we just consider the verbose flag a debug log level setting?

mule/mule.py
```diff
@@ -0,0 +1,90 @@
+import os
```
bricerisingalgorand, line 1: Is the purpose of this module to be the implementation of our cli options/args? It seems like it's only doing a few bits and pieces of that. I agree that breaking up driver makes sense, but i'm not totally sure i see the logical separation here

mule/util/yaml/env_var_loader.py
```diff
@@ -1,22 +1,18 @@
...
 def path_constructor(loader, node):
     value = os.path.expandvars(node.value)
     if env_var_path_matcher.match(value):
-        cprint(
-            messages.CANNOT_EVALUATE_ENV_VAR.format(node.value),
-            'yellow',
-            attrs=['bold']
-        )
-        return ''
+        logger.debug(messages.CANNOT_EVALUATE_ENV_VAR.format(node.value))
```
bricerisingalgorand, line 14: I think this should be under logger.warning, since we want to let the user know in case not including the var is a mistake

mule/logger.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def setup_file_logger(logger, tempfile):
+    file_handler = logging.FileHandler(tempfile)
+    file_handler.setLevel(logging.DEBUG)
+    file_formatter = ColoredFormatter(
+        log_format,
+        datefmt=log_date_format,
+        reset=True,
+        log_colors=colors
```
bricerisingalgorand: Will these colors work in most text editors? I think they may not in the more modern ones

mule/logger.py
btoll: Yes

mule/main.py
btoll: > Should we be using logger here instead of print?

I think `print` functions are fine for the `main` function.  If we were to use the logger, it would be writing to `stderr`, which isn't what we want.

btoll: > Also should we just consider the verbose flag a debug log level setting?

I initially was doing it that way, but I decided against it because there's a difference between verbosity and what information should be captured for a debugging session.

mule/mule.py
btoll: This module is the public API.  We want it separate from `main.py`, which is only intended to be parsed for cli operations.

mule/util/yaml/env_var_loader.py
btoll: I think there's a case for that and I considered it.  If we were to increase the priority to `warning`, then it would always print, which is what it currently does and impedes tooling.  Setting it as `debug` is a lower priority and will only print when in debug mode, which is an opt-in.

I think a case can be made that this should only be displayed for debugging purposes, since there viable use cases where the env vars won't be defined by the time that this code runs, both when `mule` is used as a cli tool and is imported as a dependency.

mule/logger.py
btoll: That's an implementation detail of the `colorlog` package.  I took that from its github page: https://github.com/borntyping/python-colorlog

Why do you think text editors wouldn't support them?

mule/logger.py
bricerisingalgorand: I think those options should be separated. It would be rare that I want to log to a file

mule/main.py
bricerisingalgorand: Hmm I think that makes sense in the context of a cli

mule/util/yaml/env_var_loader.py
bricerisingalgorand: What do you guys think @onetechnical @egieseke ? I think it would make more sense to log it as warning by default, and maybe have an ignore warnings option

mule/logger.py
bricerisingalgorand: I just tested it and it does mess the logs up some. Also it looks like the formatter isn't getting added to the handler in this function.

vscode:
```
[2021-01-08 13:47:30] INFO     test         Logs for this session are available in /tmp/mule.2021-01-08T18:47:30.701396.log[0m
[2021-01-08 13:47:30] DEBUG    test         --------------------------[0m
[2021-01-08 13:47:30] DEBUG    test         2021-01-08T18:47:30.701396[0m
[2021-01-08 13:47:30] DEBUG    test         --------------------------[0m
```

vim
```
[2021-01-08 13:47:30] INFO     test         Logs for this session are available in /tmp/mule.2021-01-08T18:47:30.701396.log^[[0m
[2021-01-08 13:47:30] DEBUG    test         --------------------------^[[0m
[2021-01-08 13:47:30] DEBUG    test         2021-01-08T18:47:30.701396^[[0m
[2021-01-08 13:47:30] DEBUG    test         --------------------------^[[0m
```

bricerisingalgorand:
> > Command completion will be helpful. Log changes and refactoring of file names should probably be in separate PRs.
> 
> Separate PRs is a good idea, when we get better at timely reviews I'll start doing that more.

Smaller PRs are easier to review quickly though

mule/logger.py
btoll: It's a nice feature.  Debug logging is an opt-in feature specified by the `--debug` flag, and as we grow out our debugging information, it will be nice to be captured somewhere.  Also, it's logging to `/tmp` so it's not dirtying anyone's workspace.

There have been plenty of times that I have been working with a tool and wish that I had known the cli magic that had caused a flow through an application.  Having these logs is nice, and if someone never looks at it, who cares, they are small files that are cleaned up by the OS.

mule/util/yaml/env_var_loader.py
btoll: We could have a logger context to lookup an option like that.  That would be better than having a pass-through variable.

I think it makes sense either way, but if we go for the warning and a context object, I think that should be included for when I do the job logger (which I have a ticket for).

btoll:
> > > Command completion will be helpful. Log changes and refactoring of file names should probably be in separate PRs.
> > 
> > 
> > Separate PRs is a good idea, when we get better at timely reviews I'll start doing that more.
> 
> Smaller PRs are easier to review quickly though

It doesn't seem to matter the size of my PRs.  Once I see they get reviewed quickly, I'll start doing it more.

Also, I had a smaller PR for the last mule ticket, and it caused a lot of confusion (the iterative approach) so I wanted to try doing it a different way this time by getting most of the work done in one ticket.

mule/logger.py
btoll: I'm not seeing that on Debian and konsole, depends on your terminal.

bricerisingalgorand: Mac is a pretty popular os for development. I think we're better off disabling color when writing to a file

btoll: It's the terminal, not the OS.

But, yes, I can disable the color.

btoll: Fixed.

bricerisingalgorand (approved):

egieseke (approved):

PR #66, opened 2021-01-08
Add test suite for public API

btoll:
## Summary

Added test suite for public API.  It's not comprehensive, but it's a good start.

## Test Plan

`pytest`


tests/mule.yaml
```diff
@@ -0,0 +1,89 @@
+#!mule
+agents:
+  - name: deb
+    dockerFilePath: docker/build/cicd.ubuntu.Dockerfile
+    image: algorand/go-algorand-ci-linux-ubuntu
+    version: scripts/configure_dev-deps.sh
```
onetechnical, line 6: These tests are very algorand-specific. We should probably make general use case tests.

tests/mule.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  - name: rpm
+    dockerFilePath: docker/build/cicd.centos.Dockerfile
+    image: algorand/go-algorand-ci-linux-centos
+    version: scripts/configure_dev-deps.sh
+    buildArgs:
+      - GOLANG_VERSION=`./scripts/get_golang_version.sh`
+#    env:
+#      - NETWORK=$NETWORK
+#      - VERSION=$VERSION
```
onetechnical: Nit: remove comment

mule/main.py
```diff
@@ -23,6 +23,7 @@ def _get_job_config(mule_config, job):
     tasks = job_def.get("tasks", [])
     mule_agents = mule_config.get("agents", [])
     task_configs = []
+    agents = []
```
onetechnical, line 26: Was the absence of this a bug?

tests/test_mule.py
```diff
@@ -0,0 +1,64 @@
...
+    (mule_configs, "hello", ["deb"]),
+    (mule_configs, "echo", []),
+    (mule_configs, "double_dip", ["deb", "rpm"]),
+], indirect=["mule_configs"])
+def test_list_env(mule_configs, job, agent):
+    env = mule.list_env(mule_configs, job)
+    assert set(env.keys()) == set(agent)
+    if len(agent):
+        agent_keys = {key for ag in agent for key in env.get(ag).keys()}
+        assert agent_keys == set(["NETWORK", "VERSION"])
```
onetechnical, line 46: Hmm, as noted above, I'm not wild about putting go-algorand-specific details into here. Maybe shorten and generify them in the test config?

onetechnical (changes requested):
Some thoughts on this. I'd prefer to dig out some of the go-algorand-specific stuff from the test suite.

mule/main.py
btoll: Yes, it surfaced when writing the tests.

tests/test_mule.py
btoll: Will do.

tests/mule.yaml
btoll: Will do.

tests/mule.yaml
btoll: I'll remove the algorand mentions.

btoll: Actually, looking at this more closely, it will be a lot of work to remove the algorand bits.  We'll either need to push up new images that will then be referenced from this yaml (and that will still have "algorand" it it as it would be in one of our docker repos) or it will have to be built on-the-fly every time, which will add considerable time to running the tests.

It's not leaking any sensitive information, too.  These images are publicly-available.

tests/test_mule.py
btoll: All of this is public information, both visible in s3 and on github.

mule/main.py
btoll: Done

tests/mule.yaml
btoll: Done

tests/mule.yaml
onetechnical: I'm not as concerned with information leakage here as that it's a super-specific use case for mule that people outside of our company wouldn't understand.

tests/test_mule.py
onetechnical: As before - seems like it's testing our specific algorand implementation, not testing mule itself, which is why it rings odd to me.

onetechnical (dismissed):
I'm going to mark this as approved because the core of it seems fine, but it does feel like we're testing the algorand implementaiton (with the algorand-specific bits) than mule itself, which may be confusing to third parties.

tests/test_mule.py
btoll: How is it not testing mule?

tests/mule.yaml
btoll: How is it super-specific?

egieseke: I agree that the test should be independent of algorand, since Mule is intended to be a generic pipeline utility.

tests/test_mule.py
egieseke: It would be best to structure the mule tests to not depend on algorand specifics.

tests/mule.yaml
btoll: Again, how is this super-specific?  Can either one of you elaborate?

egieseke (commented):
Try to make test generic.  We can add additional test to go-algorand-ci that test using mule for the algorand pipeline.

btoll:
> Try to make test generic. We can add additional test to go-algorand-ci that test using mule for the algorand pipeline.

Again, it would be helpful if either one of you can elaborate on how these tests aren't general.

tests/mule.yaml
egieseke: this reference for example: `image: algorand/go-algorand-ci-linux-ubuntu`

btoll: But that's just a publicly-available image.  We could test any image.  I chose this because there's a good chance we already have that cached.

You want to look at how the public API is being tested, which is completely independent of the image used.

btoll: I did address this in one of my previous comments on this thread.

btoll: Also, the way that `mule` was designed makes it difficult to use just any generic image without having knowledge of the way the project was structured.

This ticket will address that: https://github.com/algorand/go-algorand-internal/issues/951

onetechnical: The entire agent config is very algorand-specific. This isn't a hiding information thing, it's test case thing. I believe the test is that we set an env in the agent, so stripping down the agent to a simple docker container with an env of like ENV_VAR_1 and ENV_VAR_2 would be more readily understandable.

egieseke (dismissed):

onetechnical (approved):
I think this is fine for now, because it's testing the core thing we want (env var from agent).

For the test case I probably would have fabricated something simpler to be more explicit, (like maybe use the hello-world docker agent and pass an env there?). Otherwise people might think NETWORK and VERSION are special variables we're testing against.

bricerisingalgorand (approved):

PR #68, opened 2021-01-12
Added pre-commit hooks

btoll:
## Summary

Added two git pre-commit hooks.  One is for `pytest` for the unit tests and the other is for `pycodestyle`, a PEP8 linter.  Installing is not mandatory, but it's a nice-to-have.

If installed, the git hooks can be bypassed when committing (for instance, in the case that there are too many linter warnings):
```
git commit --no-verify
```

I know they work because I'm using them.

### DIFF
--- a/scripts/hooks/install.sh
+++ b/scripts/hooks/install.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-echo "$(tput setaf 2)[INFO]$(tput sgr0) Adding pre-commit hooks to .gitconfig..."
-git config --global --add hooks.pre-commit.mule "pycodestyle.sh"
-git config --global --add hooks.pre-commit.mule "pytest.sh"
+echo "$(tput setaf 2)[INFO]$(tput sgr0) Adding local pre-commit hooks to .git/config..."
+git config --local --add hooks.pre-commit.mule "pycodestyle.sh"
+git config --local --add hooks.pre-commit.mule "pytest.sh"
 
--- a/scripts/hooks/install.sh
+++ b/scripts/hooks/install.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+echo "$(tput setaf 2)[INFO]$(tput sgr0) Adding pre-commit hooks to .gitconfig..."
+git config --global --add hooks.pre-commit.mule "pycodestyle.sh"
+git config --global --add hooks.pre-commit.mule "pytest.sh"
+
--- a/scripts/hooks/pre-commit.d/pycodestyle.sh
+++ b/scripts/hooks/pre-commit.d/pycodestyle.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+if ! which pycodestyle > /dev/null
+then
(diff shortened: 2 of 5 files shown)


btoll:
This is an easy way to run tests: https://github.com/algorand/mule/pull/65

scripts/hooks/install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo "$(tput setaf 2)[INFO]$(tput sgr0) Adding pre-commit hooks to .gitconfig..."
+git config --global --add hooks.pre-commit.mule "pycodestyle.sh"
+git config --global --add hooks.pre-commit.mule "pytest.sh"
```
onetechnical: Should these be global? 🤔

scripts/hooks/pre-commit.d/pycodestyle.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+if ! which pycodestyle > /dev/null
```
onetechnical, line 3: We should add a recommendation to install pycodestyle and pytest to the README.

scripts/hooks/pre-commit.d/pytest.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 0
+fi
+
+FILES=$(git diff-index --cached --name-only HEAD 2> /dev/null | grep ".py\b")
+
+if [ -n "$FILES" ]
+then
+    echo "$(tput setab 7)$(tput setaf 4)[INFO]$(tput sgr0) Running $(tput bold)pytest$(tput sgr0) pre-commit hook..."
+
+    cd tests || exit
```
onetechnical, line 15: Also add a note about how to write tests to the README.

onetechnical (changes requested):
Mostly want to ensure we have documentation around this.

Did we decide we wanted to enforce PEP8? I'm not 100% on this yet; I don't like the tabbing recommendation, but that's just me. Haven't reviewed the rest of it yet.

scripts/hooks/pre-commit.d/pytest.sh
btoll: There's lots of good documentation about `pytest` out there.

btoll:
> Mostly want to ensure we have documentation around this.
> 
> Did we decide we wanted to enforce PEP8? I'm not 100% on this yet; I don't like the tabbing recommendation, but that's just me. Haven't reviewed the rest of it yet.

My feeling is that this is an easy win for those who want to add the hooks.  It's optional, so it won't be enforcing anything.

We've been talking about adding these checks for a while, I wanted to act on it in a way that's unobtrusive.

scripts/hooks/install.sh
btoll: Good idea, I changed to local (`.git/config`).

scripts/hooks/pre-commit.d/pycodestyle.sh
btoll: Updated README.

onetechnical (approved):

egieseke (approved):

PR #69, opened 2021-01-22
Remove empty values from `env` block

btoll:
## Summary

This is a blocker.

When recipes or cli env vars contain empty values, i.e. "FOO=", "", the docker task will throw an exception.  This is because the following isn't valid syntax in a `docker run` command:
`docker run -e ""` etc.

## Test Plan

Tested manually and via unit tests.

### DIFF
--- a/mule/task/docker/__init__.py
+++ b/mule/task/docker/__init__.py
@@ -39,9 +39,9 @@ class Docker(ITask):
         build_args_str = ""
         tags = [image]
         tags_str = ""
-        if build_args is not None and len(build_args) > 0 :
+        if build_args is not None and len(build_args) > 0:
             build_args_str = f"--build-arg {' --build-arg '.join(build_args)}"
-        if tags is not None and len(tags) > 0 :
+        if tags is not None and len(tags) > 0:
             tags_str = f"-t {' -t '.join(tags)}"
         docker_command = f"docker build {build_args_str} {tags_str} -f {self.machine['dockerFilePath']} {self.machine['context']}"
         subprocess.run(docker_command.split(' '), check=True)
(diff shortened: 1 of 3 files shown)


egieseke (approved):

mule/task/docker/__init__.py
```diff
@@ -122,17 +122,20 @@ def run(self, image):
...
-        for volume in self.check_block_format(self.machine['volumes'], ":"):
+        for volume in self.format_block(self.machine['volumes'], ":"):
             docker_command.extend(['-v', volume])
 
         docker_command.append(image)
         docker_command.extend([self.machine['shell'], '-c', self.command])
         atexit.register(self.kill, container_name)
         subprocess.run(docker_command, check=True)
 
+    def trim(self, block):
```
onetechnical, line 136: Suggestion: if we don't use trim elsewhere, I'd just incorporate this into the format_block routine.

onetechnical (approved):
Had a suggestion, not a requirement.

Random thought: is there ever a case where we'd want to pass an env as empty, as in unsetting a variable? I don't think we in particular have any use cases, but there may be larger more general cases.

mule/task/docker/__init__.py
btoll: I had initially thought that I'd test this function which is why I had broken it out.  But, I'll probably fold it back into `format_block` on subsequent work.

btoll:
> Had a suggestion, not a requirement.
> 
> Random thought: is there ever a case where we'd want to pass an env as empty, as in unsetting a variable? I don't think we in particular have any use cases, but there may be larger more general cases.

My initial thought on is that wouldn't be a `mule` issue, we'd want to make sure that it was unset in the calling environment.

PR #70, opened 2021-02-03
Radical new poc

bricerisingalgorand:
## Summary

This implements a new cli called muleci that can do generic build/publish/deploy features of services we need to get up and running.

## Test Plan

I used it to build/publish/deploy both of the dispenser servers and i've used the mule cli to do most of our pipeline release process to ensure that it still works.


egieseke (approved):
I tested this by using it to deploy the new dispenser.

onetechnical (approved):
This is a super lot to review, but nothing jumps out to me as particularly egregious. Good that we've validated against a real deployment. This big caveat is that this is missing A LOT of documentation. If you could put in a ticket to document this (especially setting up a new service), that would be helpful.

I think we also discussed how the Npm archetype might be better as a Docker archetype or something; we could discuss that at sysarch or maybe create a ticket for that as well.

PR #73, opened 2021-03-11
Update ingress hostname to support overrides

bricerisingalgorand:
By default the application_name is used but now it supports a basename override which will change the first subdomain of the address.


algobarb (approved):

scripts/push.sh
```diff
@@ -6,11 +6,7 @@ VERSION=$1
 
 if [[ -z $VERSION ]]
 then
-    VERSION=${1:-$(sed -n 's/.*=\s*''\([0-9.]*\)''/\1/p' ./mule/__init__.py)}
-    # Strip single quote from beginning.
-    VERSION="${VERSION#\'}"
-    # Strip single quote from end.
-    VERSION="${VERSION%\'}"
+    VERSION=$(python3 setup.py --version)
```
onetechnical, line 9: In general I like this better, but it also pre-supposes the binary is called as python3. I think that's always okay?

onetechnical (approved):
I think this is okay...provided it was tested of course!

scripts/push.sh
bricerisingalgorand: True. I think most python installers set that up by default. The main reason I did it this way though is osx by default makes "python" python 2, so this would help prevent this breaking when doing releases on mac.

PR #75, opened 2021-03-17
Feature/support multiple apps under same dns basename

bricerisingalgorand:
Support multiple apps under same dns basename.

Missed committing some file_util changes needed to get the dependency installs on linux to work. Removed the helm repo update it was doing since we don’t need it and it fails on machines that don’t have helm repos set up.


onetechnical (dismissed):
Seems fine, though I did not test.

